Issue #65, opened 2020-03-28
Keyswap service doesn't work immediately on boot.


Jonchun:
**Install Type:** Bare Metal
**Distro:**  Elementary OS 5.1 Hera
**DE:** Pantheon
**Branch:** master

I've noticed that the keyswap stuff doesn't work immediately on first login/boot, and requires me to open up either Firefox or Terminal (elementary terminal) before it starts recognizing the keyswaps. I tried opening up/focusing Telegram (autostarts on boot) and SublimeText3 (doesn't autostart) but it doesn't do anything. However, once I focus/open either Firefox or Terminal (the only ones I've found so far), the keyswap service starts working immediately.

The key combination I'm using to test is cmd+space (alt_key+space) which pulls up Albert.

```
cat /tmp/kinto/caret
cat: /tmp/kinto/caret: No such file or directory
```
**journalctl**
```
-- Logs begin at Fri 2020-03-27 09:03:30 EDT, end at Fri 2020-03-27 22:43:12 EDT. --
Mar 27 21:38:06 jchun-desktop bash[1400]: ls: cannot access '/home/jchun/.config/ibus/bus': No such file or directory
Mar 27 21:38:06 jchun-desktop bash[1400]: cat: /home/jchun/.config/ibus/bus/: No such file or directory
Mar 27 21:38:06 jchun-desktop systemd[1382]: Started keyswap for Terminal and GUI.
Mar 27 21:38:06 jchun-desktop bash[1400]: Importing user_config.json...
Mar 27 21:38:06 jchun-desktop bash[1400]: Data from user_config.json imported successfully.
Mar 27 21:38:06 jchun-desktop bash[1400]: Running init command: gsettings set org.gnome.mutter overlay-key ''
Mar 27 21:38:06 jchun-desktop bash[1400]: Starting keyswap...
Mar 27 21:38:06 jchun-desktop bash[1400]: First window name: none
Mar 27 21:38:06 jchun-desktop bash[1400]: gui: none
Mar 27 21:38:15 jchun-desktop bash[1400]:      TelegramDesktop
Mar 27 21:38:20 jchun-desktop bash[1400]:      Sublime_text
Mar 27 22:42:16 jchun-desktop bash[1400]:      TelegramDesktop
Mar 27 22:42:21 jchun-desktop bash[1400]: firefox: Firefox
Mar 27 22:42:22 jchun-desktop bash[1400]: file /tmp/kinto/caret does not exist
[...]

PR #70, opened 2020-03-30
Change ibus install condition logic

Jonchun:
Since I'm comfortable with Python, I thought I could at least help out by taking a peek at the setup code and fixing an unofficially documented issue in #65 where ibus doesn't get setup on initial `./setup.py`.

I think the previous logic was incorrect in checking for the existence of `homedir + '/.config/ibus/bus'`. Instead, we want to create the local ibus parent directories since the goal is to prevent `ls: cannot access '~/.config/ibus/bus': No such file or directory`.

The issue where it doesn't run is likely not unique to Elementary OS, and as far as I can tell, going to affect at a minimum all Ubuntu-based distros because `apt-get install ibus -y` doesn't seem to create that directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,8 @@ if len(check_ibus) == 0:
 if runpkg != 0:
 	requirements(pkgm)
 
-if os.path.exists(homedir + '/.config/ibus/bus') and cmdline("ls ~/.config/ibus/bus -1rt") == "":
+os.makedirs(homedir + "/.config/ibus/bus", exist_ok=True)
+if cmdline("ls ~/.config/ibus/bus -1rt") == "":
 	install_ibus()
 
 


rbreaves:
The running of ibus-setup should create that directory & the files that belong in it. (If that doesn’t execute then the logic may need to slightly change to ensure that it does when ibus & its daemon is not or has not been ran.)

Jonchun:
You currently have it so `install_ibus()` runs `ibus-setup`. However, you don't ever run `install_ibus()` because you're checking to see if that directory exists here:
```
if os.path.exists(homedir + '/.config/ibus/bus') and cmdline("ls ~/.config/ibus/bus -1rt") == "":
	install_ibus()
```
If it automatically gets created via `ibus-setup`, I think we can just change it to

```
if cmdline("ls ~/.config/ibus/bus -1rt") == "":
	install_ibus()
```

Jonchun:
A lot of these potential issues/confusion can also likely be avoided by using [Pathlib](https://docs.python.org/3/library/pathlib.html). If you're open to it, I don't mind rewriting your `setup.py` script to be more Pythonic.

PR #88, opened 2020-04-02
Add kitty to terminal emulators list

indrekj:
See: https://sw.kovidgoyal.net/kitty/


rbreaves:
Looks good, I appreciate the contribution 😊.

PR #107, opened 2020-04-11
Add Caps Lock to Escape modifier

indrekj:
There already is "swapescape" which swaps caps lock with escape. I however don't use Caps Lock at all, and so far have bound it to Escape on all my systems. I think this can be useful for others as well. As it's only a couple of lines of code, I decided to put PR up.

### DIFF
--- a/.xkb/symbols/mac_gui
+++ b/.xkb/symbols/mac_gui
@@ -4,6 +4,10 @@ xkb_symbols "swapescape" {
     key <ESC>  { [ Caps_Lock ] };
 };
 hidden partial modifier_keys
+xkb_symbols "caps_escape" {
+    key <CAPS> { [ Escape ] };
+};
+hidden partial modifier_keys
 xkb_symbols "caps_shiftlock" {
     replace key <CAPS> { [ Shift_Lock ] };
     modifier_map Shift { Shift_Lock };


rbreaves:
Works for me. I don't really use Caps at all either.

jwickens:
How does one configure this from a clean install of kinto?

rbreaves:
Depends, it is in the tweaks menu now of the app or system tray on linux. On Windows though you would need to uncomment the relevant lines in the config file and restart the program. A tweaks menu needs to be built out for Windows still, but everything else should be pretty similar btwn the 2.

Windows
~/.kinto/kinto.ahk

Re-run
~/.kinto/kinto-start.vbs

jwickens:
I'm on ubuntu 20. I can see the tweaks menu but not the system tray. I've tried the caps lock tap is escape, hold down is ctrl, but I'm often too slow for the tap. I use vim in vscode and I need to exit a lot, and this often translates to mashing escape + another key at the same time. I'd like just caps lock to escape modifier, but I'm not really understanding the config files well.

rbreaves:
Disable the remap in the tweaks menu and add the following then.

Starting with line 38 then

~/.config/.kinto/kinto.py
```
define_conditional_modmap(lambda wm_class: wm_class.casefold() not in terminals,{

    Key.CAPSLOCK: Key.ESC,   # Caps2OnlyEsc
```

Restart the service and it should be pretty easy to make whatever changes you need from that single file and restart the service as needed.

As far as the system tray goes that is due to the direction Ubuntu is moving in - but there are appindicator trays you can install still. Supposedly 2 different ones but I have only gotten one to work. The note is mentioned after the initial install, and I think in the readme as well as old tickets I am sure. Beyond that I can't really give much direction on the system tray issue.

I prefer to install and use Ubuntu Budgie these days as it tends to be the only distro that is 99% ready to go out of the box for me without any weirdness going on. 2nd to that Ubuntu Mate, and Enso OS are nice as well as a desktop OS.

jwickens:
Perfect! Thank you for the help. I was so close to coming up with that config on my own but didn't quite understand this file.

Thank you also for the tip on Budgie will check it out.

PR #122, opened 2020-04-25
Close streams and check fopen return in check_caret

fratajczak:
Previously, kintox11 would segfault after a while at line 101 (fseek) because fopen returned NULL with errno 24: Too many open files in check_caret.
This commit prevents fseek from crashing by checking the return value of fopen and prevents fopen from failing by closing the streams after we are done using them.


rbreaves:
Thanks for the patch! I have merged into master will make a new official release.

PR #180, opened 2020-05-18
Stop ctl-h (physical) / Super-h (soft) minimising windows

Mahoney:
It's not an OS/X shortcut, and it conflicts with IntelliJ's shortcut for type
hierarchy.


setup.py
```diff
@@ -216,7 +216,7 @@ def setShortcuts():
 			cmdline('kwriteconfig5 --file "$HOME/.config/kglobalshortcutsrc" --group "kwin" --key "Walk Through Windows Alternative" "none,none,Walk Through Windows"')
 			cmdline('kwriteconfig5 --file "$HOME/.config/kglobalshortcutsrc" --group "kwin" --key "Maximize Window" "none,Meta+PgUp,Maximize Window"')
 			cmdline('kwriteconfig5 --file "$HOME/.config/kglobalshortcutsrc" --group "kwin" --key "Window Maximize" "Meta+Ctrl+F,Alt+F10,Maximize Window"')
-			cmdline('kwriteconfig5 --file "$HOME/.config/kglobalshortcutsrc" --group "kwin" --key "Minimize Window" "Meta+h,Meta+PgDown,Minimize Window"')
+			cmdline('kwriteconfig5 --file "$HOME/.config/kglobalshortcutsrc" --group "kwin" --key "Minimize Window" "Meta+PgDown,Minimize Window"')
```
rbreaves, line 219: Can you change  the "Minimize Window" "Meta+PgDown,Minimize Window" to be --key "Minimize Window" "Alt+F9,Meta+PgDown,Minimize Window" instead?

I need to always be binding to some sort of universal hotkey between all OS's otherwise the kinto.py file will get very messy. Super+h was what a determined I would make the default. Additionally if KDE only uses 1 hotkey I believe proper convention is to set one of them to "none".

rbreaves:
No longer in use, Kinto now uses whatever the default it without making modifications (only additions when nothing is available at all).

PR #182, opened 2020-05-18
Make toggle-application-view shortcut more mac like

Mahoney:
Toggle-application-view defaults to `<Super-a>` (virtual)/`<Ctrl-a>` (physical)
This clashed with using `<Ctrl-a>`(physical) for move cursor to beginning of
line (OS/X).
`LaunchB` seems the closes OS/X analogue - FN-F4 on a mac keyboard.


rbreaves:
In terminals this would normally be a none issue - but in GUI based apps where such a wordwise input could supported then this would make sense. I am wondering if it was part of my wordwise setup in the xkb edition for the GUI side already and accidentally got left off during the transition.

I will merge this, but I believe I will be needing to remap Ctrl (physical) + a to the Home key and Ctrl (physical) + e to the End key.

PR #196, opened 2020-05-31
- Added Windows 10 Emoji Panel Ctrl-Cmd-Spacebar

rbreaves:


rbreaves:
Additional fix too.. for some reason the Sublime text 3 remaps for VS Code started giving issues as it was uncommenting commented in code when it shouldn't have been.

PR #252, opened 2020-08-03
- System Tray for Ubuntu, Ubuntu Budgie & Mate.

rbreaves:


rbreaves:
Removed from master manually. Did not intend to immediately add this to master. It is now in dev and will sit there for a few days before moving into master.

PR #258, opened 2020-08-11
Fixing Alt + Backspace on Windows

matteron:
Just a quick fix for the autohotkey script so that `ctrl + backspace` is only accepted if it's a "raw" input from the keyboard to prevent `alt + backspace` from triggering it and causing you delete the whole line when you just wanted to delete a word.


rbreaves:
Thanks!

Issue #277, opened 2020-09-24
Script error if non-numeric characters are entered


Yuribtr:
**Is your feature request related to a problem? Please describe.**
When setup.py start to run, user prompted to enter choice value, which can be non-numeric if in case of user's mistake. In this case script is falling to error.

**Describe the solution you'd like**
To make soft and user-friendly reaction on mistake.

**Describe alternatives you've considered**
Message that no valid number was entered or return to beginning of script.

**Additional context**
This will my first pull-request ever. I will prepare fix ASAP.

rbreaves:
The setup has now been completely rewritten w/ full GUI and new features added. Please try out the v1.2 release.

PR #278, opened 2020-09-24
Make user-friendly reaction on mistake when entering non-numeric symbol

Yuribtr:
Make user-friendly reaction on mistake when entering non-numeric symbols during setup. Fixes issue #277


rbreaves:
I appreciate it, but no longer needed due to the complete rewrite of the setup.

PR #282, opened 2020-09-30
Improved Docs

dhiraj-upadhyay:


rbreaves:
I believe this may have actually been a form of spam due to some type of contest that was occurring around this time.

PR #283, opened 2020-10-01
update F string in place of format

shanmohd:


rbreaves:
Thanks, I'll run those changes a bit later and see about merging. I am surprised anyone has even really looked at that file tbh. It sorta leaked out of my alpha branch months ago by accident when I was merging other changes. I am not sure if I will be using that file any or not, at least not for a new setup config like what was originally intended due to other changes I have made.

Kinto will soon have a system tray app for practically every major distro and a regular app shortcut could be added to launch it as well. Once that occurs asking setup questions will be most irrelevant, people can make their changes as needed after the install and quickly. They'll know if it works or not, although I could give the option to run the keycheck script just to make sure the expected key combinations are being made. Would be interesting if I could also let the test hotkeys for various apps as well in the same terminal based script or in the new GUI interface. If nothing else that would give people a good idea of what all is supported.

rbreaves:
I will be removing the keycheck file as it has been superseded by a similar function written for an actual GUI based app and it was never in actual use. Thanks for looking into it though.

PR #284, opened 2020-10-02
add proper instruction to remove the application from linux

MayankMaheshwar:
Record the installed files in files.txt and then remove.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,8 @@ sudo apt install python3
 To Uninstall Kinto
 
 ```
-./setup.py
+python3 setup.py install --record files.txt (record install files)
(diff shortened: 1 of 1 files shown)


rbreaves:
Interesting approach, it won’t undo all shortcuts modified though & if that removes the shortcut backups then you could actually interrupt an uninstall from resetting os shortcuts back to normal.

I plan to undo the few system level shortcuts I modify though & relocate that into Kinto.py fully. It was written that way for simplicity when Kinto used xkb vs xkeysnail & prior to the heavy usage of regex during setup.

rbreaves:
The installer has changed, there are no real changes to hotkeys that occur anymore unless no hotkey exists at all for a particular function. Closing the pull request.

PR #292, opened 2020-10-16
Update String

henrychase-linux:
Make strings that are important


rbreaves:
This’ll be coming very soon but I need to finish a couple of things.. a new & very short wizard after install & a remaking of the installer & uninstaller.

Most everything else I think is done.. but I might should also run more tests on other distros again before release. I do most of my dev work on Ubuntu Budgie.

PR #326, opened 2020-11-18
send f1-12 keys to terminal from mac keyboard

trin4ik:
kinto is great work, thank you!
by default mac keyboard send FN keys to terminal, when pressed F1-F12. its bullshit, if you work with Midnight Commander, etc.


rbreaves:
This will be a difficult thing to implement and merge as everyone is going to have a different media key layout on their F/FN keys. A better route will be to add another screen or screens to my setup wizard under Linux to properly map media keys to their F key counter parts assuming a user would want that across the board or only in the terminal. I will also want to implement something similar on the Windows side otherwise the two will diverge.

Thank you for the PR but currently the concerns I mentioned above will prevent me from merging it as is. I am glad you were able to remap them however and will refer to this in the future.

trin4ik:
ok, you know better ) cool, if it will be is in setup, or comment in config file.

fracture91:
[It's worth mentioning that you can tweak this behavior at the OS level and avoid depending on kinto for it](https://askubuntu.com/a/7553).  I imagine Windows has something similar?

PR #354, opened 2020-12-11
Fixes dev tools shortcut. Closes #347

### LINKED ISSUE #347: Cmd+option+i to open developer tools in Firefox
Sorry don't mean to bombard issues but I've just switched from Mac to Elementary OS and I'm trying to write down all shortcuts I'm trying to hit from memory and that don't work with Elementary + Kinto.

On Mac, I used to open developer tools with cmd+option+i which I think is the only way to do it. On elementary+kinto I have to hit cmd+shift+i to open them.

rbreaves:


rbreaves:
Ok, I have tested the latest commit, pretty confident it is all working now, so committing changes into master. Makes me realize how much better xkeysnail could handle typing words.

rbreaves:
Just to note.. the alt+home method and then typing in preferences could work - but I prefer settings to go into a new tab, not sure how macOS does it. I opted to do Cmd+t to open a new tab and then type in the preferences location.

Of course the initial method I had also involved the alt menu system and triggering settings from there - but I prefer this implementation as there is no opportunity for the Alt menu to appear at all - at least on windows. Never tested it under linux but I assume the same holds true.

PR #359, opened 2020-12-12
Add Hyper terminal to Linux

albertofem:
Recently stumbled upon this tool to use the Mac keyboard under Linux and Windows. It works great!

Hyper (https://hyper.is/) is already added for Windows, not so for Linux!


rbreaves:
I probably just didn't want to bother with installing Hyper under linux given that it is an electron app (would likely be a 100mb+ install), but I do appreciate you testing and adding it. Will merge!

PR #385, opened 2021-02-04
Make Command+Num on linux jump to tabs in browsers

fracture91:
See "Go to Tab 1 to 8" and "Go to Last Tab":

https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly#w_windows-tabs

See "Jump to a specific tab" and "Jump to the rightmost tab":

https://support.google.com/chrome/answer/157179?co=GENIE.Platform%3DDesktop&hl=en#zippy=%2Ctab-and-window-shortcuts

I tested this on Ubuntu with an Apple keyboard on Chromium and Firefox, but I'm not sure how it behaves on the rest of the things in `browserStr`

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -209,6 +209,15 @@ define_keymap(re.compile(browserStr, re.IGNORECASE),{
     K("RC-Q"): K("RC-Q"),           # Close all browsers Instances
     K("M-RC-I"): K("RC-Shift-I"),   # Dev tools
     K("M-RC-J"): K("RC-Shift-J"),   # Dev tools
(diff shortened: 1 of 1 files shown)


rbreaves:
I had been delaying on merging this due to yes, not yet testing against various browsers or checking it against Windows. It appears that the hotkeys for chrome is not fully documented as Alt does indeed work vs Ctrl.. and Alt tends to work more often than Ctrl on Linux so it is perfect for that platform.. Windows however may do both or just Ctrl, so leaving it alone on the Windows side is actually preferrable.

All that to say - I will accept this PR as is because there does not appear to be any changes needed before accepting it. Thanks!

PR #386, opened 2021-02-05
Fix Command + Shift + Grave same app switching so it moves backwards, not forwards

fracture91:
On Ubuntu 18 I noticed that I can press Command + Grave and switch between windows belonging to the same app, but holding Shift didn't make them move backwards.  This might have just been a copypasta mistake from the line above?


rbreaves:
Looks like a copy pasta error & thanks!

PR #412, opened 2021-03-06
temporary fix for using kinto with Synergy

m000:
If you use kinto and [Synergy](https://symless.com/synergy) at the same time, kinto will apply transformations when Synergy has the focus. This messes up input because the transformed keypresses are consumed by a different desktop than the one kinto is running on.

However it is not possible to write a rule to avoid transforming keypresses specifically for Synergy, because Synergy doesn't seem to set the wm_class.

The  conditional modmap in this commit solves the problem. But it will also affect any other applications not setting the wm_class. For this, the modmap is commented by default, and Synergy users (which shouldn't be too many) need to enable it manually.

 The fix is considered temporary. I have contacted the Synergy team separately, explaining the issue and asking them to set an appropriate wm_class.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -44,6 +44,12 @@ define_multipurpose_modmap(
     {                                         # Placeholder
 })
 
+# Fix for avoiding modmapping when using Synergy keyboard/mouse sharing.
+# Synergy doesn't set a wm_class, so this may cause issues with other
+# applications that also don't set the wm_class.
+# Enable only if you use Synergy.
+# define_conditional_modmap(lambda wm_class: wm_class == '', {})
+
 # [Global modemap] Change modifier keys as in xmodmap
 define_conditional_modmap(lambda wm_class: wm_class.casefold() not in terminals,{
 


rbreaves:
Interesting, I wonder if there are any other possible indicators that synergy has focus than the typical wm_class method. Even if so then it is  a question of bring that into xkeysnail/python.

m000:
I'm not intimate with Synergy code and Xlib, so I'm not sure if the have simply omitted to set a wm_class or they can't have one because of the way the software is implemented.

If wm_class can be set, then it's Synergy's omission not setting it, even if workarounds can be implemented in kinto/xkeysnail.

Having said that, I've also checked the xkeysnail code that retrieves the wm_class. It doesn't look super-clean: When Xlib wrapper returns None, the value is converted to the empty string, which is then used to evaluate kinto's conditionals.

rbreaves:
Having written c code for handling xlib output for wm_class myself it would be surprising to hear that the wrapper would be clean tbh. Xlib can return such unexpected results at times it can feel like your implementing a hack on top of a hack on top of a hack to avoid states that’ll bring it crashing down.

I was very surprised to find a python wrapper like xkeysnail working as well as it does, not sure how much of that is due to the author or the author that wrote the xlib python library.

PR #422, opened 2021-03-13
Fix "toggle maximization state" shortcut for gnome/kde

RedBearAK:
Changed "LC" to "Super" to fix the "toggle maximization state" shortcut. [ K("RC-Super-f"):K("M-F10"), ]
Not changing the other three instances, not sure if they need to be changed. (eos, manjaro, manjaro)


rbreaves:
Oh, thanks I meant update that after you mentioned it and forgot lol.

PR #426, opened 2021-03-14
Fix start script symlink path on Windows install

jfsso:
The symbolic link creation for the start script at the startup directory was actually trying to create the link in a wrong directory and failing. It couldn't start itself later at line 84 because the symbolic link to the startup script did not exist at its expected place.

I believe this solves the issue reported in rbreaves#256.


rbreaves:
Odd, under Windows 10 that path has consistently worked for me. Visually that change looks valid so I will accept it w/o testing. I’m pretty familiar w/ windows paths so I’d be very surprised if it broke anything

If anything my Windows 10 versions must have more virtual paths than some versions of Windows.

PR #429, opened 2021-03-18
Fix for installing Kinto on Ubuntu 20.04.x and 20.10.x

RedBearAK:
- Moved the installation of pip to just prior to the first attempted invocation of "pip3 install pillow" at line 212 in original file.
- If Ubuntu, trigger install of package "gcc", needed to complete Kinto setup. 
- Split "ubuntu" from "debian" and "linuxmint" to trigger install of additional package "python3-setuptools" after "python3-dev". Might also be necessary for Debian and Mint in the near future.


RedBearAK:
Closing this and replacing with merged PR #437

PR #432, opened 2021-03-18
Ensure git is installed before trying to use it.

RedBearAK:
Git is needed to show Kinto version at start of setup. Therefore, ensure that git is installed in the case that the user is installing from a manually downloaded ZIP file rather than a `git clone`. 

EDIT: Of course I mixed up shell code with python code. This won't work. I have to figure out how to do it in python. Duh.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,6 +121,9 @@ if os.path.isdir(homedir + "/.config/kinto") == False:
 	os.mkdir(homedir + "/.config/kinto")
 	time.sleep(0.5)
 
+if ! [ -x "$(command -v git)" ]; then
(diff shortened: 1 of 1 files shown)


rbreaves:
There's a few ways, examples of running shell code through python is sprinkled throughout the python setup file I am sure. I am sometimes not even consistent on which method I use, sorta depends on the command - some encapsulations work better than others at times.

RedBearAK:
Tried some ways of running it with `cmdline` but it never worked. Took it out because the problem is really just superficial in `setup.py`. Not being able to display the version and build number. Not a big deal. The fatal errors that would terminate setup were in `xkeysnail_service.sh` and that was cured by installing `git` there. Good enough for now. 

No hurry, in fact probably a good thing I've had a day to figure out how I screwed up the first time.

PR #440, opened 2021-03-29
Add shortcuts for browser prev and next tab switching

mamercad:
I use "cmd-[" and "cmd-]" frequently, and I'm probably not alone.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -234,6 +234,8 @@ define_keymap(re.compile(browserStr, re.IGNORECASE),{
     K("RC-Key_7"): K("M-Key_7"),
     K("RC-Key_8"): K("M-Key_8"),
     K("RC-Key_9"): K("M-Key_9"),    # Jump to last tab
+    K("C-Left_Brace"): K("C-Page_Up"),
+    K("C-Right_Brace"): K("C-Page_Down"),
 })
 
 # Open preferences in browsers


rbreaves:
Sounds good to me, I have some additional shortcuts for back and forward based on Cmd-Left/Right - but it is caught up in my dev branch atm as I need to do some more work on the installer to get that to setup right fully before merging. This shouldn't create any conflicts though so will merge as is.

mamercad:
> Sounds good to me, I have some additional shortcuts for back and forward based on Cmd-Left/Right - but it is caught up in my dev branch atm as I need to do some more work on the installer to get that to setup right fully before merging. This shouldn't create any conflicts though so will merge as is.

Hey, thanks for making this -- my muscle memory thanks you :smile:

RedBearAK:
@mamercad @rbreaves 

Cmd+Braces on macOS browsers is by default used for page (back/forward) navigation. This merged PR interferes with that default action. 

Cmd+Shift+Braces on macOS browsers works for tab (left/right) navigation. Therefore this is the more "correct" shortcut to be using for tab nav under Kinto's Mac-style keyboard remapping. 

PR #536 leaves the lines to map Cmd+Braces to Ctrl+PgUp/PgDn for tab nav, but they are disabled by default, as they should be for most users. If this PR is merged, you can re-enable the lines manually if you want to continue using Cmd+Braces for tab nav. 

PR #536 also adds Cmd+Shift+Braces for tab nav, so Linux browsers will function just like macOS browsers. 

PR #541 adds similar functionality for Cmd+Shift+Braces to do tab nav in all supported Linux file managers for consistency with web browsers and the Finder tab nav behavior.

PR #441, opened 2021-03-29
Reformat WM_CLASS lists as vertically stacked for easier reading/editing

RedBearAK:
The python lists of WM_CLASS names are some of the longest lines in the `kinto.py` file, exceeding the recommended line length for python, and making it more difficult than necessary to scan through visually and add new entries without accidentally creating duplicates or making syntax errors. 

No functional problems seem to come from rearranging the lists as vertical stacks with one WM_CLASS entry per line. 

Lists are also -re-arranged alphabetically.


RedBearAK:
This PR also adds support for QTerminal (LXQt default terminal).

PR #443, opened 2021-03-29
Remove duplicate line for mapping Ctrl+V in terminals

RedBearAK:
There is an identical line a few lines below where Z/X/C/V are all mapped together.


rbreaves:
Sure enough. I guess xkeysnail might ought to do what autohotkey does or allow for a strict type mode to avoid duplicate remaps. Thanks.

PR #448, opened 2021-04-01
Adds Cmd-Dot sigint to Terminals

rbreaves:


RedBearAK:
@rbreaves 

The Cmd+Dot mappings look good. Thanks!

But I tried the "fix" that I see here in the PR, and wound up having a very strange inconsistency problem with Alt+Tab. Like, sometimes it's on Alt (Ctrl) and sometimes it's on Win (Alt) and it doesn't even seem consistent between terminals and GUI apps. 

I took the lines out and everything went back to normal. I don't have any more details at the moment. 

Removed: 

```
mscodes.extend(remotes)
```

Added: 

```
remotes.extend(terminals)
remotes = list(set(remotes))
```

I'm in KDE Neon at the moment. 

I tried commenting out the second line like this, but everything was still weird until I removed both: 

```
remotes.extend(terminals)
# remotes = list(set(remotes))
```

RedBearAK:
@rbreaves 

Yeah, so it's Alt+Tab (Ctrl+Tab) from a GUI app to a terminal, then I have to Win+Tab (Alt+Tab) in order to get out of the terminal. But other shortcuts I've tested are behaving normally.

rbreaves:
You, right.. I see that this was intentionally compounding and I have now broken that :/. I will need to change this.

rbreaves:
It is fixed on master now.

And yea, I see what I did wrong now pretty clearly. Reverted some of the code near the top back to what it was and then broke out your RC-Dot change in the remotes block to a separate "not in terminals" block that actually includes remotes. This way the compounding effect of some GUI based hotkeys can still properly impact the terminal still as it was doing before.

I think the order of the apps is just fine - it is layered though and can be hard to grasp initially I grant you that (obviously because I just screwed up lol). The approach though is to be properly inclusive though and to be explicit as little is possible.

RedBearAK:
@rbreaves 

I see what you did. Kind of the opposite of what I was doing. 

Don't fully understand it yet but I patched the relevant lines into my `kinto.py` and everything seems to be behaving as expected now, including the terminal/GUI Cmd+Dot mappings. 

Looks good. See you next time.

PR #450, opened 2021-04-01
Finder mods for  Linux file managers (squashed commits)

RedBearAK:
Finder mods with commits squashed into a single edit. 

Conflict with the other was probably just because you made a minor edit to the existing Nautilus block (which this entirely replaces, if accepted). But this is an edit of the new file in master, so no conflict.


RedBearAK:
Closing in favor of PR #463.

PR #451, opened 2021-04-01
Add zypper support, make unipkg.sh output easier to read

RedBearAK:
Removing "--allow-downgrades --allow-remove-essential --allow-change-held-packages" as a replacement for "--force-yes" because none of those options are really recommended. They can potentially break the system entirely, and shouldn't actually be necessary in any normal situation when installing packages. 

This adds some strategic blank lines and "descriptions" of what is happening so things are far more clear in the terminal during install. Aids in troubleshooting which parts go wrong. 

Adds support for the zypper package manager used by openSUSE Tumbleweed.

### DIFF
--- a/linux/system-config/unipkg.sh
+++ b/linux/system-config/unipkg.sh
@@ -1,15 +1,31 @@
 #!/bin/bash
 
 if pkgmgr="$( which apt-get )" 2> /dev/null; then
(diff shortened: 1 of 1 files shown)


RedBearAK:
Closing. Superseded by PR #460.

PR #452, opened 2021-04-01
Supplemental X11/Xorg environment check

RedBearAK:
The basic check in `setup.py` failed on openSUSE Tumbleweed (GNOME Wayland login). This uses a different environment variable that may be more reliable.

### DIFF
--- a/xkeysnail_service.sh
+++ b/xkeysnail_service.sh
@@ -8,6 +8,42 @@ distro=$(awk -F= '$1=="NAME" { gsub("[\",!,_, ]","",$2);print $2 ;}' /etc/os-rel
 typeset -l dename
 dename=$(./linux/system-config/dename.sh | cut -d " " -f1)
 
+typeset -l sessiontype
+sessiontype=$XDG_SESSION_TYPE
+
+# Supplemental check to see if user is in a verifiable X11/Xorg environment. 
+# Show warnings, ask user if they want to continue installing if X11/Xorg can't be verified. 
+if [ "$sessiontype" == "wayland" ]; then 
+	echo 
+	echo "  ==================================================== "
+	echo "  Seems like you may be running Wayland, not Xorg/X11. "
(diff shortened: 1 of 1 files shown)


xkeysnail_service.sh
```diff
@@ -8,6 +8,42 @@ distro=$(awk -F= '$1=="NAME" { gsub("[\",!,_, ]","",$2);print $2 ;}' /etc/os-rel
...
+	echo "  ==================================================== "
+	echo  
+elif ! [ "$sessiontype" == "x11" ]; then 
+	echo 
+	echo "  ==================================================== "
+	echo "      Make sure you're in an Xorg/X11 environment.     "
+	echo "     Kinto/xkeysnail doesn't work yet under Wayland.  "
+	echo "  ==================================================== "
+	echo 
+elif [ "$sessiontype" == "x11" ]; then
```
rbreaves, line 30: This looks repeated, so the 2nd elif won't ever trigger.

RedBearAK: One is using "!" so testing for NOT finding "x11" after the first "if" tests for finding "wayland". 

The second `elif` tests for positively finding "x11". 

So the middle one will be triggered if the result is "" (blank due to variable not present or empty) or "unknown" (which I have seen but maybe in a different variable) or anything else besides either "wayland" or "x11". This makes the warning message not claim to have detected Wayland, but just state that it hasn't successfully detected X11/Xorg. 

There's no "else" because I couldn't think of a reason to have a "testless" default condition in this circumstance. 

As far as I know all three conditions are working when appropriate.

RedBearAK: I guess it could be rearranged so the middle `elif` becomes an `else` at the end with no condition, but I wanted to first explicitly deal with the states of "wayland matched" or "x11 not matched" before declaring success at the end when matching "x11". Either way it should work.

PR #453, opened 2021-04-01
Ask user to verify if Manjaro is updated

RedBearAK:
Install will fail (GUI window will not appear) on Manjaro unless system is fully updated first. This patch asks users to verify that they've run "sudo pacman -Syu" before attempting install. 

Failure is possibly related to the GLIB error mentioned later in the install script. Not certain. But a full update always got Manjaro to finally show the GUI installer window.


rbreaves:
pip3 was really too far up the script. I moved it down in an earlier commit after this was created, but I have also merged this.

PR #454, opened 2021-04-01
Inline export DISPLAY on openSUSE Tumbleweed

RedBearAK:
OpenSUSE Tumbleweed has an error where install completes but the xkeysnail service won't run due to "display not found". Double quotes together indicates display environment variable inside the command is nonexistent or empty. 

This patch inlines `export DISPLAY=` [ insert actual DISPLAY variable value] at the beginning of the main command in the xkeysnail service file, just before the file is copied into its destination. 

Don't know of any other way to fix the issue on Tumbleweed. Tried moving the "Environment" line before the "ExecStart" line in the service file, that of course made no difference.


rbreaves:
I think there is more issues than just this with tumbleweed, at least for me and my KDE install vte has issues. Working it out now.

RedBearAK:
Closing. Superseded by PR #460.

Issue #435, opened 2021-03-20
[QUESTION] Targeted distros for testing? 


RedBearAK:
@rbreaves 

While I'm doing all this testing to improve the compatibility of the installer script, I thought I would ask what your priorities are as far as which distros to focus on for testing and "officially" supporting. This is my list, in order of what I'll try to focus on first. 

I know there is no good reference for which distros are actually used more frequently. But I figure a combination of any of the top 30-ish on DistroWatch that are actually desktop-oriented distros, plus whatever gets mentioned most frequently online, is a good start. Manjaro/Arch, Fedora and openSUSE Tumbleweed get mentioned quite a lot everywhere I go. So I think they are close behind the Ubuntus for number of users. 

Ubuntu:
Ubuntu Desktop (stock edition)
Linux Mint 
Ubuntu variant editions (Kubuntu, Xubuntu, Lubuntu, Ubuntu Budgie, etc.)
Pop!_OS 

Fedora (has moved to Wayland as default)

Arch:
Manjaro - official desktop variants (KDE, GNOME, Xfce)
Manjaro - (comes in at least a dozen different desktop variants)

Planning on testing the "minimal" desktop variants whenever I get a chance to be sure the installer script pulls all necessary packages for each platform. Or at least is able to tell the user which packages are missing.

krswin:
Thanks for great work. 
Please include official support for Linux Mint Cinnamon.
Thanks.

RedBearAK:
@krswin 

Kinto is working well on every official desktop environment for Linux Mint. Applying the Xfce tweaks to the `kinto.py` file allows it to work simultaneously with Cinnamon, MATE and Xfce on the same machine, if desired. 

It's one of the most popular distros for new or less technical users so it will always be part of testing for me. I've been using Mint myself quite a lot over the past year.

krswin:
@RedBearAK Thanks a lot for reply.

rbreaves:
Lately I've been spending more time in Windows 10 than I'd like, but when I do drive linux on my laptop I boot into Ubuntu Budgie and only test the others in Virtualbox. I do now have a shared partition on my drive that allows me to continue my VM dev testing under macOS, Ubuntu Budgie and Windows without missing a beat. It is limited in size though, 200gb. So I keep some but not all VMs on there.

[...]

PR #455, opened 2021-04-01
Proposal for new xkeysnail installer

prosac:
Hi @rbreaves,

(and also @RedBearAK if you are interested ;-) )

I made a draft for how I would reorganize the xkeysnail installer code in a way that it follows _polymorphism over conditionals_.

Advantages

* No conditional clutter all over the install and uninstall procedures
* No additional cognitive load on the implementor of a distro of version specific installer caused by things that might be needed on other variants
* Clear separation into cohesive steps
* No further conditional clutter for desktop environment specific things by applying the same patter "on the inside"
* "Distro specialists" can implement the commands in the way that is most idiomatic for the distro of their choice
* Choise of language to implement the specific things in

Disadvantages

* More code
* Potential duplicate code
* More indirection

In my opinion the advantages clearly outweight the disadvantages here. The kind of duplication the approach creates is actually of the good kind. It is made straight forward and explicit what happens on which distro and version combination. 

Actually I would love to see this on python level and in the end have as few parts implemented in shell script as possible. I guess in python exist cool installer libraries that implement the command pattern in a transactional up and down manner.

Please tell me what you think.

### DIFF
--- a/linux/installers/fedora33.sh
+++ b/linux/installers/fedora33.sh
@@ -1,0 +1,185 @@
+#!/bin/bash
+#
+# The Fedora 33 specific installer for xkeysnail
+#
+# I would rather implement this using either a more elegant language like ruby or python
+# but I did not want to go too far for now either, so i sticked with reorganizing the
+# existing code and keeping only the parts applicable for fedora, since that is what I
+# use most of the time and thus have the best knowledge about.
+
+# Find out the desktop environment name using the existing means
+declare -l dename
+dename=$(./linux/system-config/dename.sh | cut -d " " -f1)
+
+echo "Desktop environment is $dename"
+
+# Fail when it is now gnome for now. All supported desktop environments should be
+# explicitly added here when the specific functions for them are also implemented
+[[ $dename == "gnome" || $dename == "kde" ]] || echo "Only gnome and KDE supported at the moment. Exiting." && exit 1
+
+# External interface
+#
+# These two functions must be provided by every distro and version specific implementation
+
+function install {
+	echo "Running intallation for Fedora 33"
+	prepare_for_$dename
+	install_common_components
+	prepare_dirs_and_files
+	do_something_i_do_not_understand_to_the_kinto_py_file
+	perform_actual_installation
+	finish_for_$dename
+}
+
+function uninstall {
+	echo "Running unintallation for Fedora 33"
+	# TODO: implement the tear down
+}
+
+# "Private" functions
+
+function install_common_components {
+	if ! [ -x "$(command -v pip3)" ]; then
+		echo "Will need to install python3-pip..."
+		sudo dnf check-update
+		sudo dnf install -y python3-pip
+	fi
+
+	if ! [ -x "$(command -v python3-config)" ]; then
+		echo "Will need to install python3-devel..."
+		sudo dnf install python3-devel
+	fi
+}
+
+function prepare_dirs_and_files {
+	mkdir -p ~/.config/kinto
+	echo "$(git describe --tag --abbrev=0 | head -n 1)" "build" "$(git rev-parse --short HEAD)" >~/.config/kinto/version
+	yes | cp -rf ./linux/kinto.py ./linux/kinto.py.new
+	yes | cp -rf ./linux/gui/ ~/.config/kinto/
+	yes | cp -nrf ./linux/initkb ~/.config/kinto/initkb
+	yes | cp -rf ./linux/killdups.sh ~/.config/kinto/killdups.sh # TODO: There must be a better way for what this does btw.
+	yes | cp -rf ./linux/trayapps/appindicator/kintotray.py ~/.config/kinto/kintotray.py
+	yes | cp -rf ./linux/trayapps/appindicator/kintotray.desktop ~/.config/kinto/kintotray.desktop
+	yes | cp -rf ./linux/gui/kinto.desktop ./linux/gui/kinto.desktop.new
+	yes | cp -rf ./linux/trayapps/appindicator/icons/kinto-color-16.svg ~/.config/kinto/kinto-color.svg
+	yes | cp -rf ./linux/trayapps/appindicator/icons/kinto-color-16.svg ~/.config/kinto/kinto-color-48.svg
+	yes | cp -rf ./linux/trayapps/appindicator/icons/kinto-invert-16.svg ~/.config/kinto/kinto-invert.svg
+	yes | cp -rf ./linux/trayapps/appindicator/icons/kinto-solid-16.svg ~/.config/kinto/kinto-solid.svg
+	yes | cp -rf ./linux/trayapps/appindicator/icons/kinto.svg ~/.config/kinto/kinto.svg
+	yes | cp -rf ./linux/xkeysnail.service ./linux/xkeysnail.service.new
+	sed -i "s/{username}/$(whoami)/g" ./linux/xkeysnail.service.new
+	sed -i "s#{homedir}#$(echo "$HOME")#g" ./linux/xkeysnail.service.new
+	sed -i "s#{homedir}#$(echo "$HOME")#g" ~/.config/kinto/kintotray.desktop
+	sed -i "s#{homedir}#$(echo "$HOME")#g" ~/.config/kinto/gui/kinto-gui.py
+	sed -i "s#{homedir}#$(echo "$HOME")#g" ./linux/gui/kinto.desktop.new
+	sudo mv ./linux/gui/kinto.desktop.new /usr/share/applications/kinto.desktop
+	sed -i "s#{xhost}##g" ./linux/xkeysnail.service.new
+	sed -i "s#{systemctl}#$(\which systemctl)#g" ~/.config/kinto/xkeysnail.desktop
+	sed -i "s#{xhost}#$(\which xhost)#g" ~/.config/kinto/xkeysnail.desktop
+	sed -i "s#{homedir}#$(echo "$HOME")#g" ~/.config/kinto/xkeysnail.desktop
+	sed -i "s/{displayid}/$(echo "$DISPLAY")/g" ./linux/xkeysnail.service.new
+	sed -i "s#{sudo}##g" ./linux/xkeysnail.service.new
+}
+
+# TODO: Find out what this is. Some search and replace i guess, but what and why?
+function do_something_i_do_not_understand_to_the_kinto_py_file {
+	perl -pi -e "\s{4}(# )(K.*)(# SL - .*fedora.*)/    \$2\$3/g" ./linux/kinto.py.new >/dev/null 2>&1
+}
+
+# The actual installation process.
+# This should be sliced into generic parts we can reuse for other distros and versions
+# and the distro specific things
+function perform_actual_installation {
+	mv ./linux/kinto.py.new ~/.config/kinto/kinto.py
+	sudo rm /etc/systemd/system/xkeysnail.service >/dev/null 2>&1
+	xkeypath="/usr/lib/systemd/system/"
+	xhost +SI:localuser:root
+	git clone --depth 10 https://github.com/rbreaves/xkeysnail.git
+	cd xkeysnail
+	git checkout kinto
+	giturl=$(git ls-remote --get-url)
+
+	if [ "$giturl" != "https://github.com/rbreaves/xkeysnail.git" ]; then
+		echo -e "\nreplacing xkeysnail with fork...\n"
+		cd ..
+		rm -rf ./xkeysnail
+		git clone --depth 10 https://github.com/rbreaves/xkeysnail.git
+		cd xkeysnail
+		git checkout kinto
+	fi
+
+	sudo pip3 install --upgrade .
+	cd ..
+
+	which xkeysnail
+	if [ $? -eq 1 ]; then
+		echo -e "\nKinto install has \e[1m\033[0;91mfailed\e[0m.\n"
+		echo -e "cd into ./xkeysnail"
+		echo -e "Run 'sudo pip3 install --upgrade .' to debug issue"
+		exit 0
+	fi
+
+	sed -i "s#{xkeysnail}#$(which xkeysnail)#g" ./linux/xkeysnail.service.new
+	sed -i "s#{xkeysnail}#$(which xkeysnail)#g" ./linux/limitedadmins.new
+	sudo mv ./linux/xkeysnail.service.new "$xkeypath"xkeysnail.service && echo "Service file added to "$xkeypath"xkeysnail.service"
+	echo "Changing SELinux context"
+	sudo chcon -v --user=$selinuxuser --type=$selinuxtype "$xkeypath"xkeysnail.service
+	sudo chown root:root ./linux/limitedadmins.new
+	sudo chown -R root:root "$xkeypath"xkeysnail.service && echo "Ownership set for root..." || echo "Failed to set ownership..."
+	sudo chmod 644 "$xkeypath"xkeysnail.service && echo "Permissions set to 644..." || echo "Failed to set permissions..."
+	sudo ln -s "$xkeypath"xkeysnail.service /etc/systemd/system/xkeysnail.service && echo "Created soft symlink..." || echo "Failed to create soft symlink..."
+	sudo ln -s "$xkeypath"xkeysnail.service /etc/systemd/system/graphical.target.wants/xkeysnail.service && echo "Created soft symlink for graphical target..." || echo "Failed to create soft symlink for graphical target..."
+	sudo systemctl daemon-reload
+	sudo systemctl disable xkeysnail
+	sudo systemctl stop xkeysnail
+	sudo pkill -f kintotray >/dev/null 2>&1
+
+	nohup python3 ~/.config/kinto/gui/kinto-gui.py >/dev/null 2>&1 &
+
+	echo -e "Adding xhost fix...\n"
+
+	LINE='xhost +SI:localuser:root'
+
+	echo -e "Kinto install is \e[1m\e[32mcomplete\e[0m.\n"
+
+	echo "If the setup wizard fails to appear then please run this command."
+	echo -e "~/.config/kinto/gui/kinto-gui.py\n"
+	echo -e "You can then either \e]8;;https://google.com\a\e[1m\e[36mG\033[0;91mo\033[0;93mo\e[1m\e[36mg\e[1m\e[32ml\033[0;91me\e[0m\e]8;;\a what dependencies you may be missing\nor \e]8;;https://github.com/rbreaves/kinto/issues/new\?assignees=rbreaves&labels=bug&template=bug_report.md&title=\aopen an issue ticket.\e]8;;\a\n"
+}
+
+# Gnome specific steps
+
+function prepare_for_gnome {
+	echo "Running gnome specific steps"
+	reset_mutter_overlay_key
+	set_show_desktop_shortcut_if_applicable
+	sed -i "s/systray = true/systray = false/g" ~/.config/kinto/initkb
+}
+
+function finish_for_gnome {
+	echo "Gnome may not support appindicators well, so by default you may need to install packages before enabling the System Tray."
+	echo "You may try one of the following extensions."
+	echo -e "    1) \e]8;;https://extensions.gnome.org/extension/615/appindicator-support/\aAppIndicator and KStatusNotifierItem Support\e]8;;\a"
+	echo -e "    2) \e]8;;https://extensions.gnome.org/extension/1031/topicons/\aTopIcons Plus\e]8;;\a"
+	echo -e "\nNote: you may want these supporting packages\n'sudo apt install gnome-tweaks gnome-shell-extension-appindicator gir1.2-appindicator3-0.1'"
+}
+
+function reset_mutter_overlay_key {
+	if [[ $(gsettings get org.gnome.mutter overlay-key | grep "''\|' '" | wc -l) != 1 ]]; then
+		bound=$(gsettings get org.gnome.mutter overlay-key)
+		echo "Overlay key, " $bound ", detected. Will be removing so Super-Space can remap to Cmd-Space for app launching.."
+		gsettings set org.gnome.mutter overlay-key ''
+	fi
+}
+
+function set_show_desktop_shortcut_if_applicable {
+	if [[ $(gsettings get org.gnome.desktop.wm.keybindings show-desktop | grep "\[\]" | wc -l) == 1 ]]; then
+		gsettings set org.gnome.desktop.wm.keybindings show-desktop "['<Super>d']"
+	else
+		if [[ $(gsettings get org.gnome.desktop.wm.keybindings show-desktop | grep "<Super>d" | wc -l) == 0 ]]; then
+			echo 'Kinto will not set your "Show Desktop" hotkey due to it already being set.\nPlease set Show Desktop to Super-D, or Edit Kinto'"'"'s config.'
+			echo "Did not run the following."
+			echo "gsettings set org.gnome.desktop.wm.keybindings show-desktop \"['<Super>d']\""
+		fi
+	fi
+}
--- a/linux/installers/ubuntu20.10.sh
+++ b/linux/installers/ubuntu20.10.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+#
+echo "Hi from the Ubuntu 20.10 installer!"
--- a/linux/xkeysnail.service
+++ b/linux/xkeysnail.service
@@ -4,8 +4,8 @@ Description=xkeysnail
 [Service]
 Type=simple
 KillMode=process
-ExecStart={sudo} /bin/bash -c '/usr/bin/xhost +SI:localuser:root && {homedir}/.config/kinto/killdups.sh && {xkeysnail} --quiet --watch {homedir}/.config/kinto/kinto.py'
-ExecStop={sudo} /bin/bash -c '{sudo} pkill -f bin/xkeysnail && exit 0'
+ExecStart={sudo}/bin/bash -c '/usr/bin/xhost +SI:localuser:root && {homedir}/.config/kinto/killdups.sh && {xkeysnail} --quiet --watch {homedir}/.config/kinto/kinto.py'
+ExecStop={sudo}/bin/bash -c '{sudo}pkill -f bin/xkeysnail && exit 0'
 Restart=on-failure
 RestartSec=3
 Environment=DISPLAY={displayid}
--- a/linux/xkeysnail.service
+++ b/linux/xkeysnail.service
@@ -4,8 +4,8 @@ Description=xkeysnail
 [Service]
 Type=simple
 KillMode=process
-ExecStart=/usr/bin/sudo /bin/bash -c '/usr/bin/xhost +SI:localuser:root && {homedir}/.config/kinto/killdups.sh && {xkeysnail} --quiet --watch {homedir}/.config/kinto/kinto.py'
-ExecStop=/usr/bin/sudo /bin/bash -c '/usr/bin/sudo pkill -f bin/xkeysnail && exit 0'
+ExecStart={sudo} /bin/bash -c '/usr/bin/xhost +SI:localuser:root && {homedir}/.config/kinto/killdups.sh && {xkeysnail} --quiet --watch {homedir}/.config/kinto/kinto.py'
+ExecStop={sudo} /bin/bash -c '{sudo} pkill -f bin/xkeysnail && exit 0'
 Restart=on-failure
 RestartSec=3
 Environment=DISPLAY={displayid}
--- a/xkeysnail_service.sh
+++ b/xkeysnail_service.sh
@@ -263,25 +263,6 @@ if [[ $distro == 'kdeneon' ]]; then
 fi
 
 if [[ $distro == 'fedora' ]]; then
-	echo "Checking SELinux status..."
-	if [[ $(perl -ne 'print if /^SELINUX=enforcing/' /etc/selinux/config | wc -l) != 0 ]]; then
-		while true; do
-		read -rep $'\nWould you like to update your SELinux state from enforcing to permissive? (y/n)\n' yn
-		case $yn in
-			[Yy]* ) setSE='yes'; break;;
-			[Nn]* ) exp='no'; expsh=" " break;;
-			# * ) echo "Please answer yes or no.";;
-		esac
-		done	
-
-		if [[ $yn == "yes" ]]; then
-			sed -i "s/SELINUX=enforcing/SELINUX=permissive/g" /etc/selinux/config
-			echo "/etc/selinux/config has been updated. Please reboot your computer before continuing."
-			exit 0
-		fi
-	else
-		echo "SELinux state should be ok for Kinto to install"
-	fi
 	if [[ $(gsettings get org.gnome.desktop.wm.keybindings show-desktop | grep "\[\]" | wc -l) == 1 ]];then
 		gsettings set org.gnome.desktop.wm.keybindings show-desktop "['<Super>d']"
 	else
@@ -427,6 +408,11 @@ fi
 
 if [[ $distro == "fedora" ]]; then
 	perl -pi -e "\s{4}(# )(K.*)(# SL - .*fedora.*)/    \$2\$3/g" ./linux/kinto.py.new >/dev/null 2>&1
+	sed -i "s#{sudo}##g" ./linux/xkeysnail.service.new
+	selinuxuser=system_u
+	selinuxtype=systemd_unit_file_t
+else
+	sed -i "s#{sudo}#`\\which sudo` #g" ./linux/xkeysnail.service.new
 fi
 
 if [[ $distro == "elementaryos" ]]; then
@@ -460,7 +446,6 @@ fi
 
 if ! [[ $1 == "5" || $1 == "uninstall" || $1 == "Uninstall" ]]; then
 	mv ./linux/kinto.py.new ~/.config/kinto/kinto.py
-	# if [ "$distro" == "fedora" ];then
 	sudo rm /etc/systemd/system/xkeysnail.service >/dev/null 2>&1
 	if [ -d /usr/lib/systemd/system ];then
 		xkeypath="/usr/lib/systemd/system/"
@@ -492,6 +477,8 @@ if ! [[ $1 == "5" || $1 == "uninstall" || $1 == "Uninstall" ]]; then
 	sed -i "s#{xkeysnail}#`which xkeysnail`#g" ./linux/xkeysnail.service.new
 	sed -i "s#{xkeysnail}#`which xkeysnail`#g" ./linux/limitedadmins.new
 	sudo mv ./linux/xkeysnail.service.new "$xkeypath"xkeysnail.service && echo "Service file added to "$xkeypath"xkeysnail.service"
+	echo "Changing SELinux context"
+	sudo chcon -v --user=$selinuxuser --type=$selinuxtype "$xkeypath"xkeysnail.service
 	sudo chown root:root ./linux/limitedadmins.new
 	# Add a check here for xkeysnail path resolving
 	sudo mv ./linux/limitedadmins.new /etc/sudoers.d/limitedadmins
--- a/xkeysnail_service_entrypoint.sh
+++ b/xkeysnail_service_entrypoint.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+#
+# This file depicts the entry point for the installation process and decides which
+# specialized set of functions to use in the overall process.
+
+# Basic functions to get distro name and version on different distros
+
+# Get distro name from /etc/os-release
+function distro_name_from_os_release_file {
+	cat /etc/os-release | grep ^NAME= | cut -d'=' -f2
+}
+
+# Get distro name from the lsb_release command
+function distro_name_from_lsb_release {
+	lsb_release -a | grep 'Distributor ID' | cut -d':' -f2 | awk '{$1=$1};1'
+}
+
+# Get version from the /etc/os-release file, just to show another variant
+function version_from_os_release_file {
+	cat /etc/os-release | grep ^VERSION= | cut -d'"' -f2 | cut -d' ' -f1
+}
+
+# Get the version on lsb distros like fedora and ubuntu
+function version_from_lsb_release {
+	lsb_release -a | grep Release | cut -d ':' -f2 | awk '{$1=$1};1'
+}
+
+declare -l distro
+distro=$(distro_name_from_lsb_release)
+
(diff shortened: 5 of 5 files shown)


rbreaves:
My initial focus was on Ubuntu/Debian based distros, but I will admit - it does get to be quite a lot to manage every distro in a single setup file. It would be a lot cleaner to break them out as you suggest. Fedora may very well be a good candidate to that first with. I will be testing your PRs out soon, but possibly not today as I try and do some serious re-organization on my end in multiple ways before I can review the outstanding PRs. (Mostly re-organizing my home office, data and devices atm.)

From the perspective of distros and repos - this would also increase the likelihood of getting that done or having others even work on getting this ported in multiple distros. I don't mind learning how to do it, but if there are clear files and clear paths to making that happen then it is probably more likely to see it happen. I am sure it would be time consuming as well to maintain multiple repos across many distributions that I might not even be testing as thoroughly as someone could.

RedBearAK:
@prosac 

I'm totally on board if it means it will be more straightforward to tackle the requirements to make each family of distros work, independently of the temperamental behavior of other distros. I've already had some difficulty interleaving new fixes for one distro among the others without accidentally breaking something else or duplicating a bunch of conditions. 

However if you try to move it into ruby or python I'll be left behind for now, since I'm mostly just familiar with shell scripting. I know it's a bit clunky but it's very linear and easy for many people to understand when your main goal is to make things happen in sequence in a terminal. 

Just looking things over...

I'm not sure I agree with a couple of things already. 

I feel like the sub-files should be more generic than distro[version]. What if you want the installer to go ahead and try to install on a newer version of Fedora than Fedora 33? Do you really want to make a file for Ubuntu 20.04 and a separate one for Ubuntu 20.10? I kind of feel like the version branching for specific oddities in new versions should happen inside the main "Ubuntu" file. 

Also, why would you want to bring a lot of the file copying/modifying stuff out of the main installer? If you ever need to change those basic functions you'd need to edit them in all the sub-files too. 

But it's quite possible I have no idea what I'm talking about. I'm just an observer.

rbreaves:
On the front of venturing out of bash or python, I guess I am ok with people using whatever language they want to use for the setup of a specific distro, but personally my preference is to switch between python and bash as I have done. I know both fairly well and generally speaking I just find doing most of the setup in bash to be simpler/faster for me and is why I use it. I'd be concerned introducing more dependencies than absolutely necessary.

If I could have done this app without any python at all I'd have done it and initially it was all bash, then c & bash, and now python & bash. Years ago I had another python implementation though that was not xkeysnail based but its performance back then was very poor compared to xkeysnail so I gave up on it then and never released it.

rbreaves:
@RedBearAK I concur with not creating entirely new setup files btwn files unless the amount of changes are significant. If the only real differences are package name changes then declare the differences early on in variables to make it clear which distro versions get what variables and then simply reference them everywhere else in the distro specific file. That completely avoids a ton of conditionals and interweaving.

RedBearAK:
@prosac @rbreaves 

Looking over these files I am suddenly wondering if it might not be far less redundant to just very clearly put different "stages" of the install sequence into separate scripts, but keep the per-distro branching within the scripts rather than going for all these different files. 

As long as each major part of the process, like installing pip, is kept to its own section of the script separate from other things, it's actually not that hard even for a coding numpty like myself to comprehend where to put in a fix for any misbehaving distro that doesn't respond well to the generic sequence. 

Unless it's done very carefully, I think going with all these separate files might actually be major overkill. I mean, think about what the installer is doing. Attempting to install a few packages, which will vary per distro but basically be the same list, and then modifying the service file and `kinto.py` based on things like detected keyboard and DE. 

It actually works surprisingly well just the way it is, with a few tweaks. 

I think that I would advocate for just a very clear reorganization and "sectionalizing" of the main script. With nice labels and such. So the right section to fix one particular problem is easily found when scrolling through. 

But again, I'm not a "coder" really, and I don't yet think the way someone who has done bigger projects might think about this.

RedBearAK:
@prosac @rbreaves 

And I think we also have to remember that while it is certainly a nicety that there is such a thorough installer available, most software like this will just list its prerequisites and expect the end user to install the necessary dependencies before trying to install something like this. 

So if the installer is a little more clear and explicit at the end after a "failure" about which packages may be missing, most Linux users at least will be able to get past that. My goal has been just to target the most popular distros that less knowledgeable users might be on, like Ubuntu and Mint and Pop_OS and a few others.

rbreaves:
Once I get my dev environment reconfigured I will be taking a closer look at all of this. Also I do agree that I don't like the killdups.sh script either.. In theory systemd I believe should be properly ending the program - but often times it leaves it running unless I run that script so that is why it exists if I recall correctly. It may just be there acting as an insurance policy or maybe some versions of systemd or distros simply behaved differently.

prosac:
@rbeaves no stress and happy re-organization!

@RedBearAK

> However if you try to move it into ruby or python

That was just a thought. Keeping it in shellscript has advantages too. You can depend on the fact that it works on any unix like system etc. Maybe when we go on like this I will make a port of a portion of the code to ruby or python just for fun and education.

> I kind of feel like the version branching for specific oddities in new versions should happen inside the main "Ubuntu" file.

Yes. That was over-optimization. You are right!

> Also, why would you want to bring a lot of the file copying/modifying stuff out of the main installer? If you ever need to change those basic functions you'd need to edit them in all the sub-files too.

This is just a draft. When we go on in this direction I would extract the generic parts too and just call them in the distro specific files.

@rbeaves

> I'd also be concerned introducing more dependencies than absolutely necessary.

You are right!

> I concur with not creating entirely new setup files btwn versions unless the amount of changes are significant. 

Yes. See above...

@RedBearAK

> Looking over these files I am suddenly wondering if it might not be far less redundant to just very clearly put different "stages"
> ...
> Normally when I have fixed something in my testing, even an oddball like Lubuntu or Manjaro LXQt, it only required a few lines here and there, or a conditional variant. Then suddenly it was working fine.
> ...

Can I play my 20+ years of programming card here? ;-) I strongly suggest to keep the things separate after the initial branching and have not a single conditional for distro switching in the actual flow. The small things here and there add up. You have to be very consequent with such design decisions. It pays out. Trust me :-)

> Unless it's done very carefully

That's how to build software :-)

> But again, I'm not a "coder" really, and I don't yet think the way someone who has done bigger projects might think about this.

I don't want to sound like a smart ass. I want to help. I work with a rather large partially quite old and polyglot code base in my day job. Refactoring to lower the friction to be able to evolve the product in a safe way is my thing.

> most software like this will just list its prerequisites and expect the end user to install the necessary dependencies 

Yeah. But this is actually very cool and part of what got me here. I like things that just work and I like helping making things just work.

> most Linux users at least will be able to get past that.

Sure, but that is no reason for me. As I said. The fact that @rbeaves when building kinto aims to provide a

> The zero effort solution.

(Quote from the readme)

got me and I would like to help to make that manageable.

@rbeaves

> In theory systemd I believe should be properly ending the program

I am sure that this is doable with just systemd means. We will find out. The less hacks are needed, the better.

> some versions of systemd or distros simply behaved differently.

Yes. Suggestion: You focus on Ubuntu and Windows, I take care for fedora. Focus is important. You can't solve everybodies problem at the same time. With my proposed restructuring the chance gets higher that somebody who runs another distro and likes to help can jump in and take that over, because there is _no_ interference with other distros. Even if that does not happen it is just better to handle this way even as a one man show.

prosac:
@rbreaves btw. i tried to buy you the Official Raspberry Pi Keyboard on adafruit as a motivational present and a thank you :-) ... but this wishlist thing does not really work. I thought I can just pay it and it automatically gets delivered to you, but now it wants my adress etc...

prosac:
And a book suggestion. One of the book every programmer should read. No matter what language he or she uses: https://martinfowler.com/tags/refactoring.html

rbreaves:
@prosac thanks for the suggested reading! And I guess I need to log into adafruit & see if I hadn’t added my address in a way that’s associated w/ the wishlist 😅. Had a similar issue w/ amazon months ago & just needed to specify my address for the list.

And it would be pretty cool to get an adafruit keyboard - they add something that’s identifying to the keyboard itself, a spec that goes mostly unused. I forget what it is though, different from vendor & product Id.

RedBearAK:
@prosac 

The very first and most important thing I would like to say is that I deeply resonate with and cherish your enthusiasm for making this project better, or making any project better. And I would never want to say anything to dampen that enthusiasm. Your contribution really helped me gain clarity on my thoughts about the best (simplest is my focus) way to make the installer better. I like thinking about such things. 

> > I kind of feel like the version branching for specific oddities in new versions should happen inside the main "Ubuntu" file.
> 
> Yes. That was over-optimization. You are right!
> 

[Dry humor] Bzzt! You have been given one demerit for saying something nice on the interwebs. [/dry humor]

> Can I play my 20+ years of programming card here? ;-) I strongly suggest to keep the things separate after the initial branching and have not a single conditional for distro switching in the actual flow. The small things here and there add up. You have to be very consequent with such design decisions. It pays out. Trust me :-)
> 

Oh yes, big projects need to be planned very well, I absolutely agree. I think mainly where we are disagreeing is the idea that the Kinto installer script would ever really grow to be so large as to be unwieldy to maintain. With proper clear organization of sections/tasks, I don't really think that will be the case unless we try to support all known distros. Which seems... unwise. 

For instance, I have an issue thread (#435) about which distros to (reasonably) target for testing the installer. Just last night I added a couple of distros after seeing what the Snap store uses for supported distro examples on the bottom of each application page on the Snap Store (snapcraft.io). There aren't that many. 

When you look at the list as just variations on the base distros of Debian/Ubuntu, Arch and Fedora/RPM-based distros, it's actually a pretty short list, and probably covers at least 90% of the type of Linux user that would ever need help getting through a glitching installer. 

> > But again, I'm not a "coder" really, and I don't yet think the way someone who has done bigger projects might think about this.
> 
> I don't want to sound like a smart ass. I want to help. I work with a rather large partially quite old and polyglot code base in my day job. Refactoring to lower the friction to be able to evolve the product in a safe way is my thing.
> 

Oh, you don't. I hope that didn't come across as sarcastic or something. I was being quite literal. (I think that's one of my basic problems on the interwebs.) The most advanced "coding" I've really done is using "functions" a few times in shell scripts. So I'm fully willing to accept the possibility that a software solution that I think of as "too complicated" could actually be a much better solution. 

Well, I guess I did create a fairly complex relational database used for production data in a seafood processing plant, but that was mostly point-and-click stuff with some relatively simple "calculations" mixed in. You could do some pretty amazing things in FileMaker by just sticking to the first few cardinal rules of relational databases. 

Where we are seeing the project differently, as above, is that I don't think the installer script will really ever need to "evolve" to be gargantuan. I know software tends to do that, and I kind of thought that about the install script a week ago, but this analytical process has changed my mind. 

I think covering at least 95% of the current desktop Linux users who can't fix the script themselves would only require enough lines to make the script (at most, ever) about 25-50% longer. And that would technically still be a pretty manageable size for a shell script. As long as it's organized and commented well enough for new people to get in there and understand what's happening and quickly apply the proper fix to get through it. (Or easily know which package or library to add manually.) A few more informative messages spit out into the terminal on failures would help most people not even need to open up the installer script. I already have several ideas about that. 

> > most software like this will just list its prerequisites and expect the end user to install the necessary dependencies
> 
> Yeah. But this is actually very cool and part of what got me here. I like things that just work and I like helping making things just work.
> 

Absolute agreement. I, too, love things that just work, but the amount of time/effort/manpower to make something like this work perfectly everywhere (rather than on 95% of user-friendly popular distros) is probably not worth it. Law of diminishing returns and all that, eh? 

Maybe I'm biased in that I've already participated in finding solutions to several installer problems on different platforms. Ben can confirm I've been a real pain in the .... over the past year. 

>> The zero effort solution.
> 
> (Quote from the readme)
> 

I could be very wrong but I the way I interpret that is it is referring to the fact that once you get it up and running it does the keyboard mapping for you without having to mess with a dozen or more different config files used by different graphics frameworks, DEs and oddball applications. In that sense, Kinto's mapping file on top of his xkeysnail fork has indeed been something that mostly "just works" for me. As I've said to Ben on occasion, there have been times lately that I've forgotten I'm not using my MacBook Pro. And that's happened to me in completely different distros with different DEs. In that sense Kinto has been the "zero effort solution" and has become basically the fist thing I download on every distro install everywhere on every machine. 

As far as the installer, my new perspective is that it actually is very well constructed with defaults that "usually" work, and just needs a few tweaks like installing a different VTE library package name on a different distro family. That particular block only has three different outcomes so far, and now works on all the distros I've managed to test. Plus knowing whether to look for "python3" packages or just "python" packages, no "3". Again, only a couple of naming variations exist. It either uses the "3", or it treats v3 packages as the default and leaves the "3" out of the package naming, only separating the older v2 packages with a "2". I don't expect that to ever balloon into a twelve-step branched tree. If a specific distro wants one or the other, you just add it to the right condition line. 

I know refactoring and working toward some vision of perfection with a piece of software is a lot of fun, and sometimes even achievable. I just don't think that's a practical vision to have for the installer file. 

Now, if you could take that same enthusiasm and tackle porting xkeysnail to be compatible with the ever-encroaching Wayland... hoo, boy... That would be neat. That's going to be hugely important in a next few years, as many users are going to be on Wayland by default (many without even knowing it). 

Thanks for the convo! Sorry if this is a big long!

rbreaves:
I've made a slight change to the installer just now if you guys want to check it out - it now downloads Kinto as a zip file, extracts and then installs it. The old method should work too - but ultimately I want this installing as easily as brew does and this update does that. @prosac @RedBearAK

RedBearAK:
@rbreaves 

I hate to say this but I don't think you can always count on `curl` being installed in a fresh distro base. I can't recall exactly which distros but I know I had a good reason to put a command to install `curl` into my installer patch for installing `pip` via the `get-pip.py` script method. 

I've also made notes already (in the "Setup fails" thread) on encountering "minimal" distro images which, astoundingly, don't even have `unzip` installed in the base software (even though they have a full desktop and web browser and other things). 

So those are two potential stumbling blocks I see for a quick "simple" install with this method, at the moment. 

I hope you'll forgive me for also pointing out that this is a much longer command that needs to be copy/pasted without making an error. The old `git` command was short enough to easily commit to memory. I have a pretty poor working memory for lengthy commands and I had that one down within three tries. The only real requirement for the old method was installing git first. Then the installer handled everything else. 

I'll certainly test it out but I guess I'm not quite sure at the moment why you would want to do it this way. The only reason I wasn't using the `git` method during all my distro testing was that I couldn't figure out how to quickly `git` my own merged patch branch easily, so I was just manually downloading the ZIP file of a specific branch. But now that I merged all my patches into master, I can `git clone https://github.com/redbearak/kinto` for testing. That seems a lot easier than trying to find this long command and copy/paste it into a terminal. 

You seem to own the domain "Kinto.sh" as a forwarding URL that brings the user to the GitHub page. What about shortening this install command by at least substituting `https://kinto.sh/linux.sh` for the `https://raw.githubusercontent.com/rbreaves/kinto/master/install/linux.sh`?

RedBearAK:
@rbreaves 

Ran into this on Manjaro LXQt: 

lxqt_no_unzip

rbreaves:
> You seem to own the domain "Kinto.sh" as a forwarding URL that brings the user to the GitHub page. What about shortening this install command by at least substituting `https://kinto.sh/linux.sh` for the `https://raw.githubusercontent.com/rbreaves/kinto/master/install/linux.sh`?

Thanks for the feedback, and the main reason is because I guess I have been lazy so far as I have not setup any host for the Kinto domain just yet. After making these changes I am more motivated to do that because https forwarding will only work IF I have an actual website with hosting setup for it. Without a host all I can do is forward http addresses which is not secure and I have decided, that for security reasons I will not instruct users to setup a script, even my own, from a none secure address. It is too easy for the redirect of HTTP to be hijacked, at least HTTPS will fail.

Also once I have hosting setup it will be presented in much the same way that https://brew.sh presents their url. Users can click a button, sadly that is not an option with github readme pages.

RedBearAK:
@rbreaves 

`gunzip` is installed on Manjaro LXQt but it can't handle multi-file ZIP files. It didn't work as a replacement/fallback for `unzip`. 

It appears there might be a way to download a tarball rather than a zip archive, using the GitHub api. 

https://gist.github.com/KazeroG/ed8c88928e5a7a3383f606546a9fa3f4

Alternately, there seems to be a way to link directly to the "latest" release file for a repo, which provides both ZIP and TAR.GZ versions. But I think you'd need to give your release files a specific file name for that to work, and name it the same way each time (no version numbers in the file name). Right now the Releases page just shows all files as "Source Code" and it wants to download the latest file as "1.2-7.tar.gz". 

You would need to always have a latest release file named "kinto-latest" or "release" or something like that, for this to work with a generic non-versioned URL: 

```
https://github.com/rbreaves/kinto/releases/latest/download/kinto-latest.tar.gz
```

With the current release naming I haven't actually been able to get that to work yet. 

But, this works (from the GitHub api notes linked above): 

```
wget --no-check-certificate https://github.com/rbreaves/kinto/archive/master.tar.gz
```

This gives me a "master.tar.gz" file, which extracts with `tar xpvf` to a "kinto-master" directory. 

So, that seems to be what you're looking for, accomplished without relying on `unzip` being installed. 

Let's see about doing it with `curl`... 

```
curl -L https://api.github.com/repos/rbreaves/kinto/tarball/master > kinto-master.tar.gz
```

This works, but unfortunately creates a directory when extracted named something like [user]-[repo]-[branch?], so not quite as clean. This one came out as "rbreaves-kinto-022dedb". Not sure at the moment if there is a way to clean that up to have it come out as a more generic name.

rbreaves:
@RedBearAK Thanks for finding that tar gz method, will definitely switch over to that and if nothing else I can just use mv to clean up the folder name. Although that SHA1 hash is useful and if there is no other way to grab that then that is useful for me. Without git I still need to get the SHA1 hash of the commit version they just downloaded and from what I saw the API does not provide that info. I was able to extract that from the zip file though.

I also use the API already to grab the release number as well so when Kinto is installed I can populate the about menu with the right release and build number.

Also this is in early stages - but I am fleshing out a new project I have codenamed Kairos atm, I already have a different & official name and new domain created for it as well. The purpose of this is for developers/programmers like myself and I will be trying pretty hard to get all of my preferred default settings into it. Right now it is relatively simple, but I expect to be putting a lot more work into it in attempt to make a "ready to go" dev environment that is fairly consistent btwn multiple popular distros. It is out of scope for Kinto though.

https://github.com/rbreaves/kairos

RedBearAK:
@rbreaves 

If it were up to me I think it'd just go with the `wget` option, leave out the "--no-check-certificate" option which seems both unnecessary and unwise, and have the equivalent `curl` command as a fallback. Or vice versa, since I think `curl` is more common as a default than `wget`. 

So I'm probably just using the `curl` options wrong, but after redirecting the output this will do basically the same thing as the `wget` version, creating a "master.tar.gz" that extracts to "kinto-master": 

```
curl -fsSL https://github.com/rbreaves/kinto/archive/master.tar.gz > master.tar.gz
```

No doubt there is a `curl` option that will do this without the redirect. 

Ah, I guess `curl` defaults to `stdout` so it's the "-o" option and the output file name. Same difference. 

Using the "api" URLs just seems like unnecessary complication. 

So a simple "which curl" and then fallback to `wget` which will be the choice on some distros rather than `curl`, should cover most distros. I hope. 

I'll definitely check the other project out. Although I will most likely not know what it would do for me at this stage.

RedBearAK:
@rbreaves 

I don't know if you want to offer this as an alternative installer URL for people who don't have `curl` installed, but this seems to successfully download and start the installer same as the `curl` based URL: 

```
/bin/bash -c "$(wget -O - https://raw.githubusercontent.com/rbreaves/kinto/master/install/linux.sh)"
```

Might be one or more other options for `wget` to fully be the same as "-fsSL" as far as being resilient to handling errors/redirects. 

Oh, before I forget, I'm a little confused about how exactly the user is supposed to input the "-r" option in this method in order to trigger the "uninstall" process. Since `setup.py` is getting initiated automatically.

rbreaves:
@RedBearAK Yea, I noticed the lack of passing arguments when working with Kairos some minutes ago, so I changed it over to use this instead. It's late, but tomorrow I will probably update the readme on kinto to a variation of what is below, because at the very end there, after the ) you can add your -r or whatever argument there np problem.

```
bash <( curl  https://raw.githubusercontent.com/rbreaves/kairos/master/linux.sh )
```

RedBearAK:
@rbreaves 

> at the very end there, after the ) you can add your -r or whatever argument there np problem.

Ah, as I suspected, that would be the only way to do it. Seems awkward, and non-obvious to the casual user. A strange thing to attach to a URL. They would definitely need explicit instructions for that. 

>The other thing to note though is the user can also go into the uncompressed kinto folder and initiate the uninstall from there.. but yea if they never really navigated to that folder to install it then they will likely be confused about how to remove it as well, assuming they don't read the readme fully.

Yep, since they are sort of bypassing that whole process if this works, they could end up pretty confused about how to reverse the process without going through the readme. 

That's another reason I'm still not a huge fan of this idea. The `git` method was very (well, relatively) straightforward. Clone, go into folder, run "./setup.py". At least the user could be expected to know where the folder is and how to get back there and run setup again to remove. 

Hey, maybe `setup.py` should actually just come out and directly ask the user (if no arguments are given) if they want to [i]nstall or [r]emove. That would probably be the lowest-friction solution for a lot of users who are adverse to reading the README. Just ask for positive input (and bypass the ask if a valid argument is present). 

Or should it be in `xkeysnail_service.sh`? I could do a patch and PR for that, no problem.

prosac:
[Dry humor] Bzzt! You have been given one demerit for saying something nice on the interwebs. [/dry humor]
😁

prosac:
> Oh yes, big projects need to be planned very well, I absolutely agree. I think mainly where we are disagreeing is the idea that the Kinto installer script would ever really grow to be so large as to be unwieldy to maintain. With proper clear organization of sections/tasks, I don't really think that will be the case unless we try to support all known distros. Which seems... unwise.

Sure. It is just a small piece of code and not the Taj Mahal of code ;-), still if you build a small house you benefit greatly from following some general patterns that make living in and with the house easier and more comfortable. I can only point to my original bullet list above again. The lever of removing all cognitive load load that comes from conditionals for n distros is simple and powerful. It is not necessary, but it would make working on the code base more fun, relaxing and straight forward (in the best sense of the word). It is a big luxury to be allowed to ignore the rest of the world, when thinking in one corner of the world. The complexity (because of distro specalties and edge cases) of this installer is high enough in my opinion to in to indicate this architectural move. (But as said above, I think with applying the pattern down to version level is too much. The small amount of extra conditional logic on distro version level is neglectable).

About the part about all known distros: I would even say that we should focus on the ones we actually use and provide well though out reference implementations for these and for now ignore the others (hard decision after already having quite a variety of distros on the list...). Adding more distro specific implementations later does not interfere with the existing ones and thus is safe and so _makes people feel save_ to contribute. That is a good feature for a code base.

prosac:
About git, curl, wget etc: I think it is ok to expect some generally available tools, even if they are not part of every fresh installation. That is general practice and does not scare people or add real friction.

Going for the "brew.sh way of installation" is a good move I think!

rbreaves:
The simplicity of a brew.sh install has always been in mind tbh, but until lately it just wasn’t my highest priority. As I am looking to polish things more it’s finally gotten some attention. But I do agree some refactoring does need to happen at this stage.

I think before splitting the install as proposed I will refactor it some - move all variables that change throughout to the top & try & set them early - & see where that leaves us. May even look at a split based on package manager instead of distro as distro flavors tend to revolve around package managers.

rbreaves:
@prosac @RedBearAK I have decided that a refactor will be happening to ensure all variables are defined and set near the top of the file. Additionally all variables will also be defined in and imported from yaml files instead.

We can continue the conversation here, but I will be closing this merge request.

An example would be this below.

Ubuntu_20.04.yaml
```
Distro: Ubuntu
DE: Gnome
Install:
    set-systemtray: s/systray = true/systray = false/g
    remove-overlay:
        overlay: gsettings get org.gnome.mutter overlay-key
        on-true: gsettings set org.gnome.mutter overlay-key ''
    # natural copy-paste - elementary only, so would actually be left on ubuntu
    remove-natural-copy-paste:
    pkg-dependencies: python3-dev
    pip-dependencies: pillow
    enable-system-layer-keys: \s{4}(# )(K.*)(# SL - .*ubuntu.*)/    \$2\$3/g
    xkeysnail-host-fix: xhost +SI:localuser:root
    final-message: echo "Gnome may not support appindicators well, so by default you may need to install packages before enabling the System Tray."
        echo "You may try one of the following extensions."
        echo -e "    1) \e]8;;https://extensions.gnome.org/extension/615/appindicator-support/\aAppIndicator and KStatusNotifierItem Support\e]8;;\a"
        echo -e "    2) \e]8;;https://extensions.gnome.org/extension/1031/topicons/\aTopIcons Plus\e]8;;\a"
        echo -e "\nNote: you may want these supporting packages\n'sudo apt install gnome-tweaks gnome-shell-extension-appindicator gir1.2-appindicator3-0.1'"
```

RedBearAK:
@rbreaves 

Well it sounds interesting. And the example YAML file you've provided looks relatively easy to understand, even for me. 

But how good would this kind of scheme be at attempting generic fallbacks when it doesn't match some exact distro/de/version condition?

RedBearAK:
@rbreaves 

FYI, `curl` is not installed in the Ubuntu 20.10 live boot "Try Ubuntu" environment, so the pasted terminal command installation technique failed. But to be fair, it does immediately tell the user how to install it. And the `curl`install works. 

But far worse than that, I think I just discovered that all those fixes I implemented for Ubuntu are unnecessary outside of the "Try Ubuntu" live environment. Probably comes down completely to the "universe" repository (where `python3-pip` can be found) not being enabled in the live boot environment by default. Your current installer actually works fine on Ubuntu after installing Ubuntu to disk, even when choosing "minimal" during install. 

So in a sense there is actually no Ubuntu installation problem. 

This is not influencing me to have the greatest of days. Although, it would be very nice if the installer reliably worked in the live environment as well as the full install, so you can test a distro on your actual hardware and with your desired keyboard shortcuts to see how well it will work with Kinto without doing an install. So, ultimately not a total waste of effort, I hope. 

But frustrating nevertheless.

rbreaves:
> @rbreaves
> 
> Well it sounds interesting. And the example YAML file you've provided looks relatively easy to understand, even for me.
> 
> But how good would this kind of scheme be at attempting generic fallbacks when it doesn't match some exact distro/de/version condition?

I am not sure why we would want a fallback and risk unpredictable behaviors but I guess if we really wanted to implement something then perhaps displaying a list of existing templates they can attempt to try if they have the same package manger. I feel like distro and de without specific versions defined is generic enough. My plan is to leave all latest LTS distros as the generic install template for any distro by having no version numbers defined.

RedBearAK:
@rbreaves 

Well, I was thinking about it like how the current installer is "usually" successful, barring a missing package on some odd distro that does things a little differently. For instance it worked just fine a few hours ago when I tried it on Zorin OS, even though I didn't really expect it to. Highly customized on the surface but just another Debian/apt based system underneath, apparently. 

There's nothing in the installer specific to Zorin, and it identifies itself as ZorinOS, gnome, zorin, and "zorin:GNOME" [distro, dename, desktopsession, currentdesktop], so only the GNOME part would actually match any condition in the Kinto installer. But since there's nothing actively blocking the install, it continues and installs perfectly (AFAICT). In other words, even though it doesn't "know" Zorin OS, it "falls back" to trying the general procedures for GNOME and an apt-based distro. And it works. 

> I feel like distro and de without specific versions defined is generic enough. My plan is to leave all latest LTS distros as the generic install template for any distro by having no version numbers defined.

That's part of what I was looking for. Not being to specific about version numbers and refusing to install on a newer version that just dropped because it's "unknown". But would this scheme make the installer say, "I don't know how to install on Zorin," rather than do what it does right now? Or can there still be a "generic" try-this fallback when the distro is not specifically recognized? Based on package manager, perhaps, because that often dictates the major package naming differences. 

Maybe I'm overthinking it, but it seems like if you don't have some sort of fallback, or perhaps more accurately a "fall through", it may necessitate far too many separate template files for every distro/DE combination. 

Probably overthinking it though. I do that.

RedBearAK:
@rbreaves 

I think we just said the same thing two different ways. LOL.

rbreaves:
@RedBearAK Yea I was kinda wondering if you hadn't understood that I was trying to say that by default my LTS based templates will not have version numbers defined - as I will assume that most users will be using the LTS builds, and that I strongly suspect that 1 version back (18.04) or the 21 releases will tend to work still. If not templates can be created to fully define their parameters.

RedBearAK:
@rbreaves 

Yeah, I think a temporary language parsing problem happened. Just terms that didn't immediately mean anything to my revving brain. I'm sure I'll be functioning better after a reboot. 

>I don't mind offering the most likely compatible template in cases of like Zorin - if the package manager and DE is otherwise supported. Warnings would be clear though that it is untested and could result in unexpected behaviors or it simply not working. 

That's what I'm talking about, basically. Most compatible template. With unknown/random distros the DE and package manager will usually point the way to success, with caveats given to the user that the outcome may be [shrug_emoji]. "You takes your chances, kid." And then they can raise an issue if it doesn't work. 

>I am not aware of any probability of breaking sudo, but I have done that before a long time ago and is one reason I like to test and not give the option to run it if I have not tested it in a VM first. I do not want someone being upset because all of a sudden they're stuck in a boot loop or can't run sudo.

The ones I remember that happening on last year were antiX and Gentoo. Bad juju. I've been meaning to maybe offer a patch to cancel the install if those are detected. Unless you did something that would keep that from happening? I'm not aware of any specific change in that direction, but I haven't messed with either of those since last year.

rbreaves:
I did work out how to fully parse yaml files in bash last night & reorganize functions so that the main function is at the top & all others can remain at the bottom & be called as needed. Many of these changes will be coming to Kinto — & Kairos will handle out of scope items & both pre & post Kinto install options.

https://github.com/rbreaves/kairos/blob/dev/yaml.sh

prosac:
> I think before splitting the install as proposed I will refactor it some - move all variables that change throughout to the top & try & set them early - & see where that leaves us. May even look at a split based on package manager instead of distro as distro flavors tend to revolve around package managers.

sounds good. iterative is the way.

regarding the yaml approach: why not nest the things in a way that the "up and down" commands are always together as pairs? so it would be very easy for the eyes to match the things and see if something can't be right.

prosac:
example for what i wrote above:

```yaml
Distro: Ubuntu
DE: Gnome
distro-version:
de-version:
Steps:
    set-systemtray:
      up: s/systray = true/systray = false/g
      down: s/systray = false/systray = true/g
    set-overlay:
      up: 
        check: gsettings get org.gnome.mutter overlay-key
        on-true: gsettings set org.gnome.mutter overlay-key ''
      down:
        overlay: gsettings get org.gnome.mutter overlay-key
        on-false: gsettings set org.gnome.mutter overlay-key 'Super'
...

```

rbreaves:
@prosac up down shouldn’t need to be broken as one can be inferred by the other. A simple & singular check ought to be enough. Also after all the work I did to get yaml to work w/ bash I’m wondering more about what you mentioned earlier, using ruby for the installer.

Considering that yaml is probably a 1st citizen of ruby & certainly popularized w/ it. Brew.sh is also pretty popular as well. I’ll have to think about it some more I guess, ultimately I want to use whatever language will encourage the most participation by the community that uses Kinto.

RedBearAK:
@rbreaves 

> https://github.com/rbreaves/kairos/blob/dev/yaml.sh

That script... is neat. In more than one sense. 

Did you see the recent announcement by Canonical that they will be using Google’s Flutter for Ubuntu apps, including the new system installer? It’s a little early but there seems to be a lot of potential there. The 2.0 release brings support for targeting native desktop applications on Windows, macOS and Linux. Perhaps you might want to look at it for at least the GUI part. 

Unless you’re planning on also porting xkeysnail to ruby, maybe it makes more sense to stick with python as much as possible and not mix too many languages in the same project? Just a thought.

rbreaves:
@RedBearAK I wouldn’t ever try porting xkeysnail to another language but I have experience w/ Ruby already - it’s a great language to use w/ yaml configs & doing automated server configurations (think Puppet). I didn’t write xkeysnail, mooz did & based it on the work of another I believe, but extended it a great deal.

The extent of using ruby would be the installer only, if I were to use it. I’m familiar enough w/ both bash & ruby imo that it wouldn’t matter to me. I’m only thinking about it in terms of potential contributors & readability as well. Looking at the popularity of Homebrew is part of why I’m saying this as well. That’s a more complicated project though by leaps & bounds.

RedBearAK:
@rbreaves 

Well, I guess the installer is a pretty much isolated "thing" in and of itself, so if ruby would really make it much easier for you to cleanly write and maintain the main file, that is a pretty strong argument to go ahead and use it. 

Just a note at the beginning, I had a couple of new variables inserted in my installer fixes. I needed to use at least one of them to positively identify Lubuntu to solve the problem of the different vte library name, because it was otherwise just identifying as "ubuntu" as far as the installer was concerned. In the example of Zorin OS they show different things, but sometimes they are the same. Depends on the distro. But they might be helpful in your identifier functions to identify variants of the main stock distros. 

```
typeset -l desktopsession
desktopsession=$(env | grep -i DESKTOP_SESSION | awk -F"=" '/=/{print $2}')
typeset -l currentdesktop
currentdesktop=$(env | grep -i XDG_CURRENT_DESKTOP | awk -F"=" '/=/{print $2}')
```

Here in KDE Neon these show "plasma" and "KDE", respectively. 

There is also XDG_SESSION_DESKTOP which seems to usually show the same thing as DESKTOP_SESSION. I don't know which is more appropriate/reliable. 

Here's Budgie 21.04: 

```
...
DESKTOP_SESSION=budgie-desktop
XDG_SESSION_DESKTOP=budgie-desktop
XDG_CURRENT_DESKTOP=Budgie:GNOME
...
```

Deepin Linux 20.2: 

```
...
DESKTOP_SESSION=deepin
XDG_SESSION_DESKTOP=deepin
...
XDG_CURRENT_DESKTOP=Deepin
...
```

And Zorin OS 15.3: 

```
...
DESKTOP_SESSION=zorin
...
XDG_SESSION_DESKTOP=zorin
XDG_CURRENT_DESKTOP=zorin:GNOME
...
```

RedBearAK:
@rbreaves 

Maybe you'd want an option on your new installer that would output what the installer "believes" the distro and DE are, so someone can get a head start on a new matching YAML template without digging into the installer code for the identifier functions to run them manually. If it comes out "wrong" from the main installer then it will be immediately obvious that part of the fix will be enhancing the identifier functions, or maybe asking the maintainers for that distro to use a unique identifier. 

Something like: 

```
./setup.rb --ident
```

And then it outputs what it thinks all these values are (maybe more?): 

```
Distro: 
DE: 
distro-version:
de-version:
```

RedBearAK:
@rbreaves 

Can we maybe open a new issue thread specifically for discussion about the installer transition/improvements? Posts on a "closed" issue like this don't seem to show up on the GitHub homepage "Recent activity" section that I see when logged in. 

I get the emails too but I also like to see the reminders of the current comment state of issues showing up in the "Recent activity" box. Helps me keep track of things. 

We'd just need to be mentioned and/or make sure we click the "Subscribe" button on the new issue thread. 

Or you could re-open this, but it's already getting pretty long.

rbreaves:
I'll just reopen it for now. Atm my plan is to possibly re-use the work for Kairos and update the installer, the bash shell component at least, to be more modular and have yaml files per distro.

prosac:
I am curious to see your approach!

A thought that came to my mind lately: basically with such an installer you are in the domain of tools like ansible, chef, salt etc. Have you considered that?

prosac:
Also babushka (which is not actively developed anymore) could be a thing: https://github.com/benhoskings/babushka.

rbreaves:
> I am curious to see your approach!
> 
> A thought that came to my mind lately: basically with such an installer you are in the domain of tools like ansible, chef, salt etc. Have you considered that?

Sorta, I have had experience with Puppet and automating raspberry pi/kiosk deployments with it in the past. Also had looked into the others you mentioned as well, but can't say I used them extensively. I also use and have created vagrants and docker images for personal and work related dev environments so working with yaml conifgs is something I am familiar with. This will be the first time that I can recall though where I will be creating a tool in both bash and python to interpret yaml.

An earlier version of Kinto was actually consuming/parsing a json config file - but the xkeysnail python based config file I think has been a step up surprisingly and yaml ought to be another step in the right direction.

RedBearAK:
@rbreaves 

Hey, I was installing Pacstall on Ubuntu 21.04 (like Arch AUR but for Ubuntu, supposedly) and noticed they use a shortened "git.io" URL in their install link. 

`sudo bash -c "$(curl -fsSL https://git.io/JfHDM)"`

Have you tried using this GitHub service to shorten your install URLs? (Don't know if it's limited to just the "github.com" domain or will work on your "githubusercontent.com" URLs.) 

Apparently you can also make "vanity" URLs by specifying the code, rather than getting back random letters from the shortener. Check this article: 

https://saraford.net/2017/02/18/how-to-use-git-io-to-shorten-github-urls-and-create-vanity-urls-049/

rbreaves:
I have not heard of git.io, I am reading it is provided by github, but I think security conscious people may be leary of it. I think I might be better off waiting until I get Kinto.sh online fully and then I can setup a proper https redirect then. I can do it now too of course but it would be http and I don't want to do that either.

RedBearAK:
Gotcha. Seems like a lot of people may not have heard of it, even though the earliest reference I found was from 2011. Just thought I'd ask. 

I guess the security-conscious types wouldn't like it even if it's supposedly restricted to GitHub URLs. There's all kinds of things on GitHub it could be linking to. And reading the article, seems like anyone could make a link to anything, not just something in their own user account. 

Got a pretty nice setup now in Ubuntu 21.04, after only a few hours. Enabled application-centric task switching, auto-hiding "menu" bar with tray indicators, auto-hiding dock. The menu and dock will even come up in "fullscreen" mode, instead of being blocked. Everything is pretty much like my typical Mac setup except the window buttons, and that can be changed easily. Still going through the process of seeing if I can get the new Fildem global app menu extension to do it's thing. If I install Albert I can probably completely forget about the excessive "Activity Overview" screen too. 

Started trying to do this last night with Fedora 34 GNOME, but was quickly reminded that the Kinto fixes for SELinux haven't been merged yet. I might mess with it again later. Will probably be moving to (or at least trying out) the new Pop_OS release when it comes out in a month or so. The previews of their new COSMIC desktop indicate they have actually listened to what most users want/use and will have some new sane defaults, like a usable "dock" out of the box, without requiring additional extensions.

rbreaves:
@RedBearAK Yea.. for me Fildem was very inconsistent on which apps it properly displayed.. all in all I think I am stucking to Ubuntu Budgie for now and they just added support for the Gnome 40 stack any ways.. from what the Budgie team says..

PR #461, opened 2021-04-05
Prompt user when no options, support more CLI options

RedBearAK:
At the beginning of the file, the "$1" argument is stored in "arg1" variable, using `sed` to allow for the user to choose the common "--" style option syntax and still produce a match. Also, the variable is declared and set to force the input to lowercase to reduce the number of test conditions needed later. 

After that, this inserts a new `if` statement, just prior to the main uninstall `if` statement, that gets triggered when there is no command-line option given. It puts up a simple prompt asking for user input on whether to [i]nstall, [r]emove or [q]uit Kinto setup. The letters (i/r/q) are chosen instead of (1/2/3) or (i/u/x) to be widely separated on the keyboard to avoid typo errors. This whole statement and user-interaction is completely bypassed if any command-line option is detected. User input causes the "arg1" variable to be appropriately set and the script continues, or the script will exit if the user chooses to quit. 

After this the normal uninstall process will proceed if the correct CLI option is found. There is no real change there except using the "arg1" variable instead of directly calling "$1". 

Two "elifs" are added to this, one triggered by the appropriate "install" CLI option (causing the last "elif" to be bypassed), and the last "elif" is triggered if "arg1" is NOT empty but wasn't validated by the earlier "if" conditions of "uninstall" or "install". This exits and shows some assistance on valid options. 

So the flow is like this: 

No CLI option? Prompt user. Set CLI option according to user choice and continue, or exit. 

CLI option? Is it "uninstall"? Then uninstall.
CLI option? Is it "install"? Then proceed to install. Ignore final "elif". 
CLI option? Is it unmatched/unrecognized? Then exit, show error and help user. 

Below is the entire script used to test the logic flow of this before I copied and pasted it in to wrap around the existing "uninstall" block of commands. You can run this in a terminal over and over again and check that it behaves as expected for any given (valid/invalid) input value (or no input). 

Also tested by copying this entire file and just inserting an "exit 0" before running any of the actual uninstall commands or proceeding into the main installer sequence. It behaves as expected within the limits of my ability to test it. 

```
#!/bin/bash

clear

# Allow users to put arguments in as "--" style options and still match
typeset -l arg1
arg1="$(echo $1 | sed -e 's/^--//g')"

# Ask user for instructions if no arguments given on command line.
if [[ $arg1 == "" ]]; then 
    while true; do
        read -rep $'\n\n  ================\n   Setup Options:\n  ================\n\n   [I]nstall Kinto\n   [R]emove Kinto\n   [Q]uit\n\n\n  (i/r/q): ' action	
        case $action in
            [Ii]* ) arg1='install'; echo -e "\n\nInstalling Kinto...\n"; break;;
            [Rr]* ) arg1='uninstall'; echo -e "\n\nUninstalling Kinto...\n"; break;;
            [Qq]* ) arg1='quit'; echo -e "\n\nExiting Kinto setup...\n"; exit 0; break;;
            * ) echo -e "\n\n Please answer [i](install), [r](remove) or [q](quit)."; sleep 1;;
        esac
    done
fi

# Start uninstall process if matching command-line argument given. 
if [[ $arg1 == "5" || $arg1 == "-r" || $arg1 == "remove" || $arg1 == "uninstall" ]]; then
    echo 
    echo "Proceeding with Kinto removal..."
    echo 
    echo "Done uninstalling."
    echo 
    exit 0
# Confirm we will continue install process if matching command-line argument given. 
elif [[ $arg1 == "-i" || $arg1 == "install" ]]; then
    echo 
    echo "Proceeding with Kinto install..."
    echo
# If command-line argument is not empty but doesn't match first two conditions, exit, display help. 
elif ! [[ $arg1 == "" ]]; then
    echo -e "\n\n  ERROR: Argument \"$arg1\" not recognized, exiting Kinto setup...\n\n"
    echo "  Valid Kinto setup script command-line arguments are: "
    echo 
    echo "  To Install: -i, install, --install"
    echo "  To Remove/Uninstall: -r, remove, --remove, uninstall, --uninstall"
    echo 
    # Exit setup because given command-line argument doesn't match any known option. 
    exit 0
fi


echo 
echo "Doing the actual install stuff..."
echo 

exit 0
```

Also included here are some proposed visually-obvious labeling for different sections of the installer script, for help finding and understanding each section when scrolling through the file.

### DIFF
--- a/xkeysnail_service.sh
+++ b/xkeysnail_service.sh
@@ -3,10 +3,25 @@
 # set about:config?filter=ui.key.menuAccessKeyFocuses
 # to false for wordwise to work in Firefox
 
+################################################################################
+################################  VARIABLES  ###################################
+################################################################################
+
+# Allow users to submit CLI arguments as "--" style options and still match
+typeset -l arg1
+arg1="$(echo $1 | sed -e 's/^--//g')"
+
+# Identify distribution
 typeset -l distro
 distro=$(awk -F= '$1=="NAME" { gsub("[\",!,_, ]","",$2);print $2 ;}' /etc/os-release)
+
+# Identify Desktop Environment
 typeset -l dename
 dename=$(./linux/system-config/dename.sh | cut -d " " -f1)
+
+################################################################################
+################################  FUNCTIONS  ###################################
+################################################################################
 
 function uninstall {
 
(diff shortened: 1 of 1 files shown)


rbreaves:
I'm hesitant to add any prompts in the CLI as I lean towards only wanting to prompt for installs, reinstalls, and uninstall from the now existing GUI. Functionality can be added to it to provide Uninstall and Re-install buttons if the program is already installed.

I don't mind adding arguments however - as long as they do not prompt the user with any questions. If questions are at play then I'd like to push that into the GUI interface.

RedBearAK:
@rbreaves 

I figured you'd be hesitant to add interaction. It would be interesting to have these choices also in the GUI. But that appears much later in the process. It can take around 15 minutes for the GUI to appear on some distros I've tested. This is right at the beginning, before any processes are shut down or packages downloaded. I don't really see how they compare. 

The interaction can be bypassed with "-i" or "install" or "--install". 

You can also, of course, pass the appropriate argument through the `linux.sh` script, right? So the download URL link can still be a clean, non-interactive setup. This would only come up when launching setup manually in the terminal with no options.

PR #463, opened 2021-04-06
Finder mods for Linux file managers (v2021-04-05)

RedBearAK:
2021-04-05 updated version of Finder mods, supporting common Linux file managers: 

```
# 
# Caja File Browser (MATE file manager, fork of Nautilus)
# DDE File Manager (Deepin Linux file manager)
# Dolphin (KDE file manager)
# Nautilus (GNOME file manager, may be named "Files")
# Nemo (Cinnamon file manager, fork of Nautilus, may be named "Files")
# Pantheon Files (elementary OS file manager, may be named "Files")
# PCManFM (LXDE file manager)
# PCManFM-Qt (LXQt file manager)
# SpaceFM (Fork of PCManFM file manager)
# Thunar File Manager (Xfce file manager)
# 
```

This adds support for the Deepin Linux default file manager, and some rearranging/code clarification improvements on the previous iteration. Closing PR #450 in favor of this update.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -291,7 +291,7 @@ define_keymap(re.compile("io.elementary.files", re.IGNORECASE),{
 
 # Keybindings overrides for Nautilus 
 # (overrides some bindings from general file manager code block below)
-define_keymap(re.compile("org.gnome.Nautilus", re.IGNORECASE),{
+define_keymap(re.compile("org.gnome.nautilus|nautilus", re.IGNORECASE),{
     K("RC-Super-o"): K("Shift-Enter"),           # Open in new window
     # K("RC-Super-o"): K("RC-Enter"),                 # Open in new tab
     K("RC-comma"): K("RC-comma"),                   # Overrides "Open preferences dialog" shortcut below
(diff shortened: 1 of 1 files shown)


rbreaves:
I've not really tested it as thoroughly as I'd have like to have done, but I made a few notes of things to look at more later. Nothing broke under my primary distro though and I didn't really see anything that made me concerned that something would break, at worse maybe nothing happens and that is acceptable. Thanks for the hard work!

Line 274
Dolphin needs - RC-Shift-o to be set for open new window during install

Line 311
spacefm - need to look at shortcut

Line 379
filemanagerStr - General file manager actions that need to be reviewed.

RedBearAK:
@rbreaves 

Thanks! Version 1.2-8 is the coolest release EVAR!!

Dolphin: Yes, needs the shortcut set before it will work. I was surprised that was not set to anything by default. 

SpaceFM is a weird one. I noted in the included comments that it even blocks Alt+Tab switching when one of its menus is open. So while I could open the menu I couldn’t activate the preferences dialog because as soon as the menu opens the key insertion sequence stops. SpaceFM seems to be set up to allow the user to change or add a shortcut to basically any menu item just by selecting the item and typing a new shortcut. And it uses Ctrl+F to create a new folder. Which is... come on, man. One of these things is not like the other. 

I can attest that while some of the shortcuts are missing or unsettable in some FMs, or simply don’t do anything, I haven’t observed anything harmful in all my testing on several different distros. As long as the “Enter key to rename” is left disabled. That’s for experts. Like me.

PR #465, opened 2021-04-06
Add lock screen shortcut

OlegSmelov:
Commented out by default because the keybinding is different in Gnome and KDE.


rbreaves:
Will need to test this, not sure it’ll apply the way you’d assume.

PR #469, opened 2021-04-07
Add PCManFM-Qt to PCManFM override, fix typos, clarify Dolphin

RedBearAK:
Just doing some sweeping up and prettifying. Main thing is adding the PCManFM fork, PCManFM-Qt to the overrides block for PCManFM. Working on the Nautilus variant reminded me.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -256,7 +256,7 @@ define_keymap(re.compile("^jetbrains-(?!.*toolbox).*$", re.IGNORECASE),{
 # Keybindings overrides for Caja 
 # (overrides some bindings from general file manager code block below)
 define_keymap(re.compile("caja", re.IGNORECASE),{
-    # K("RC-Super-o"): K("RC-Shift-Enter"),     # Open in new tab 
+    # K("RC-Super-o"): K("RC-Shift-Enter"),       # Open in new tab 
     K("RC-Super-o"): K("RC-Shift-W"),           # Open in new window
 },"Overrides for Caja - Finder")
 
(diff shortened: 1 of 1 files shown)


rbreaves:
Just to note, while I do try and keep comments aligned on the right side I also keep the indention caused by having the entire line commented out by default so that if it is enabled later it will then be aligned with the rest of it. So some off alignments are actually preferred because there's a reason for it.

RedBearAK:
@rbreaves 

... I see what you're saying. I wasn't paying attention to whether the line was commented out or not. That makes sense. Helps when visually scanning through, and lines up when you enable the line. So two more spaces if the line is disabled. 

Conforms now.

RedBearAK:
@rbreaves 

Can't get 380 and 381 to line up after submitting, but they look fine while editing in the GitHub web-UI. Even tried deleting all the whitespace and tabbing (GitHub inserts spaces) so they line up. Hit submit and it's all wonky again. 

Don't know what that's about. But not important. 

Oh, it looks fine in split view, but wonky in unified view. Weird.

rbreaves:
Looked fine in my editor and that's all I care about. Typically though if something is off it is because of a mixture of tabs and spaces. I am not a stickler on it - but a file does need to be consistent and not mixing the 2. Had some difficulty w/ a dev (not on this project) recently that couldn't get theirs configured right and it was a headache as any line they submitted could be either and then if I went in behind them my editor would properly apply the settings we had agreed to and automatically make his commits look like my commits and pretty well ruined the ability to parse out my contributions vs his.

I actually wrote a new function in bash or zsh to override "git add" so that it would always prompt me on whether or not to ignore whitespace differences so that my editor would completely fix the tabs and mask my changes vs his. It was a very ugly hack to keep git history clean though for auditing or bug tracking purposes.

rbreaves:
Also while this project, and I think python in general, wants tabs only for indentation, I prefer tabs first, and spaces after. I mean typically you have no spaces unless you add comments off to the side like we do. But this is python so spaces it is.

RedBearAK:
I see. I'll try to remember that. Logical.

rbreaves:
I apologize.. I meant to say python wants spaces only for indentation, not tabs lol. @RedBearAK

RedBearAK:
I thought something was a little off about that. Spaces it is. LOL.

PR #470, opened 2021-04-07
ctrl+fn arrow-up/arrow-down on linux browsers

dyu:
Same behavior/layout as mac when the keyboard's page-up/down keys are triggered with fn+up/down (e.g. no standalone page-up/down keys).

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -400,6 +400,8 @@ define_keymap(re.compile(browserStr, re.IGNORECASE),{
     K("RC-Key_9"): K("M-Key_9"),    # Jump to last tab
     K("C-Left_Brace"): K("C-Page_Up"),
     K("C-Right_Brace"): K("C-Page_Down"),
+    K("Super-Page_Up"): K("C-Page_Up"),
+    K("Super-Page_Down"): K("C-Page_Down"),
 })
 
 # Open preferences in browsers


RedBearAK:
@dyu 

Hi David. 

As a Mac user, I developed the remapping blocks to make file managers behave like the Finder. But I'm not familiar with this shortcut. 

What exactly is Ctrl+PageUp and Ctrl+PageDn supposed to do in the file manager? I'm trying it in Finder on my MacBook Pro and it doesn't seem to do anything different than just PageUp and PageDn (Fn+Up and Fn+Dn) by themselves. 

And for gedit, and the web browsers? What is the purpose of the shortcut? In Firefox on my Mac it's switching between browser tabs. Is that what you're expecting? It doesn't do the same in Finder, even if I load up some tabs. 

If I understand correctly the Fn key is ignored or not available to key remappers. So what is relevant is just that you're using the Page Up/Down keys. 

Also, the labels in the file manager section are named like "Overrides for Dolphin - Finder" only because they are "overriding" some shortcuts in the more general file manager remapping section at the end of the file manager group. The one that's labeled "File Managers - Finder". You'd probably want to just call yours something like "Gedit - Editor".

rbreaves:
I am confused by this as well, but will look closer tomorrow. But yes the fn key is not capable of being remapped, although the normal fn remapping can still come through because it is on the hardware layer - there's just no guarantee on what that will be.. so if we are literally trying to remap the physical Ctrl+Fn + whatever other key then this PR will likely be denied.

Fn+Arrow key may commonly become pgup or pgdn keys but I can't ensure that and I will have to boot into macOS later to see with Ctrl+fn+arrow keys do.. but even at that anything Fn related is pretty much dead in the water. I haven't even merged the PR that wants to remap the top row media keys because of the difficulty in there being no consistency up there even btwn macbooks.

RedBearAK:
@dyu @rbreaves 

It's still not doing anything for me in the Finder on macOS, but I see that here in Dolphin on KDE Neon Ctrl+PgUp or PgDn moves between tabs just like in the web browser. So I assume this is all about tab navigation. 

I'll take a look at the possibility of adding this to the general file manager block. Linux file managers already mimic other behaviors of web browsers, like Ctrl+L to enter the location bar. Have to see if it's compatible with all the other supported file managers. If it is, you won't need to put it in the override block just for Nautilus. 

Probably will look into it tomorrow.

RedBearAK:
@rbreaves 

> I am confused by this as well, but will look closer tomorrow. But yes the fn key is not capable of being remapped, although the normal fn remapping can still come through because it is on the hardware layer - there's just no guarantee on what that will be.. so if we are literally trying to remap the physical Ctrl+Fn + whatever other key then this PR will likely be denied.

I think you can ignore the Fn part, if I understand correctly he was just referring to the fact that he has to use Fn to get his arrow keys to turn into PgUp/PgDn, like on a MacBook. He seems to be just looking at remapping physical Ctrl+PgUp back to actual (logical) Ctrl+PgUp instead of Super+PgUp. For tab navigation. 

I want to look at whether this could be enabled for all file managers and then overridden if there is some conflict. But that will require testing all ten supported FMs. I can start looking at it tomorrow. Right now I'm completely knackered, as a British person might say.

RedBearAK:
@rbreaves 

I'm very tired but I'm having an inkling that this might actually belong most appropriately in the generic GUI block. Seems like there may be a lot of applications that use Ctrl+PgUp/PgDn for switching tabs. 

Anyway, I'll see if all the known file managers can handle it as soon as I can.

RedBearAK:
@dyu 

I've tried this in both KDE Neon and Fedora 34 GNOME and these shortcuts don't seem to do anything for me in gedit (Text Editor). 
```
# Keybindings overrides for gedit 
define_keymap(re.compile("gedit", re.IGNORECASE),{
    K("Super-Page_Up"): K("C-Page_Up"),
    K("Super-Page_Down"): K("C-Page_Down"),
},"Overrides for Gedit - Editor")
```

I also couldn't find Ctrl+PgUp/PgDn listed anywhere in the keyboard shortcuts popup for gedit. Are you using customized keyboard shortcuts in your gedit? 

I only see these in the default gedit shortcuts: 

```
Ctrl+Alt+PgUp/PgDn       Go to next/previous document
Shift+Ctrl+Alt+PgUp/PgDn       Go to next/previous tab group
```

dyu:
> I think you can ignore the Fn part, if I understand correctly he was just referring to the fact that he has to use Fn to get his arrow keys to turn into PgUp/PgDn, like on a MacBook. He seems to be just looking at remapping physical Ctrl+PgUp back to actual (logical) Ctrl+PgUp instead of Super+PgUp. For tab navigation.

100% this.  Ignore the `fn` I mentioned.  Linux browsers by default will navigate the tabs with ctrl+page-up/down.  Mac browsers do this too.  Since the physical ctrl key is remapped to super, this PR addresses that.

dyu:
> I also couldn't find Ctrl+PgUp/PgDn listed anywhere in the keyboard shortcuts popup for gedit. Are you using customized keyboard shortcuts in your gedit?

Did you try Ubuntu?  I'm on Ubuntu 16.04 and using gedit 3.18.x
The ctrl+page-up/down shortcuts have the same function (tab nav) on the browser, gedit, nautilius and terminal.

RedBearAK:
@rbreaves 

Until I get feedback from @dyu I won't have any idea how he's making this work in gedit, and for some reason Ctrl+PgUp/PgDn won't do anything in Nautilus for me, either with Kinto enabled or disabled, even though they are listed as shortcuts for tab switching in the keyboard shortcuts dialog for Nautilus. 

But these shortcuts do work in Firefox and Dolphin (file manager), and the same physical keys do the same thing in Firefox on my Mac. So... I don't know. It's very weird. He's putting the shortcut mappings in the Nautilus section and the browsers block. The browsers one I understand, but I just can't replicate the functions in any file managers besides Dolphin. I've tried it in Fedora 34 GNOME as well as KDE Neon. Doesn't do anything in gedit for me in either environment. 

So I can only recommend the change to the browsers block at this time. And I believe this shortcut of physical Super+PgUp/PgDn (logical Ctrl+PgUp/PgDn) should supercede the use of the Ctrl+Brace shortcuts for tab navigation. The Cmd+Brace shortcuts on my Mac do Back/Forward in Firefox, not Left/Right tab navigation. That's what I've been using the braces for here in Firefox in Linux, so it messed me up a bit when you added that recently to the browsers block. Fortunately I noticed and immediately disabled it. 

I feel like this might need a tweak option, to enable or disable braces as back/forward or tab navigation.

RedBearAK:
@dyu @rbreaves 

> > I also couldn't find Ctrl+PgUp/PgDn listed anywhere in the keyboard shortcuts popup for gedit. Are you using customized keyboard shortcuts in your gedit?
> 
> Did you try Ubuntu? I'm on Ubuntu 16.04 and using gedit 3.18.x
> The ctrl+page-up/down shortcuts have the same function (tab nav) on the browser, gedit, nautilius and terminal.

I was testing in Fedora, which is probably the closest to stock GNOME. 

I just tried Ubuntu 20.10 and can't get those shortcuts to work in Nautilus, gedit, or the GNOME terminal. But it does work in Firefox. 

The shortcuts are listed in the terminal shortcuts preferences, but are not listed anywhere in gedit's current shortcut sheet. They are listed in the Nautilus shortcut sheet. 

It doesn't seem to matter whether I have Kinto enabled or disabled. The shortcuts just don't work for me in the GNOME applications. 

I'm at a loss to understand this phenomenon.

RedBearAK:
@dyu @rbreaves 

OK, here is the story. I did a bare metal install of Ubuntu 20.10. 

In Ubuntu 20.10, the tab navigation shortcut in the current version of gedit is really Ctrl+Alt+PgUp/PgDn. 

However the normal shortcut of Ctrl+PgUp/PgDn is working as David says on Firefox, GNOME Terminal and Nautilus in this version of Ubuntu. 

So once again, I recommend adding this to the browser block while disabling the Ctrl+Brace shortcuts by default. 

And this should be a safe addition to both the Nautilus and Dolphin overrides blocks. I will have to test further on the bare metal Ubuntu install to see if it will also apply to enough other file managers to make sense to put this in the general file managers block. 

I have no idea at this point whether the issue is that I'm running KDE Neon and testing inside GNOME Boxes VMs, or some strange hardware quirk with my laptop's PgUp/PgDn keys. Nothing about it makes sense so far. It's only the GNOME applications that refuse to respond to the shortcuts. Always works fine in Firefox and Dolphin, both inside and outside the VMs. 

Ben, this could be a problem with your planned automated testing. I'm assuming you'd be trying to do that in some kind of VM environment.

RedBearAK:
@dyu @rbreaves 

I've done enough testing now that I would be comfortable at the very least adding this set of shortcuts to the general file manager block, to have it apply to all file managers. SpaceFM, as usual, is the only odd man out so far. It uses C-M-PgUp/PgDn like the newer versions of gedit. Easily dealt with. 

But if this is in the main GUI block, it works in all sorts of applications, from Firefox to file managers to terminals. I suspect it will work the same way in a large number of applications that have a "tabbed" window interface. The only question is whether it could do any harm if it was a general GUI shortcut. 

One issue is that Ubuntu has a shortcut for switching virtual desktops, which is Super+PgUp/PgDn (probably Super+Home/End too, depending on the number and configuration of virtual desktops). That gets overriden by this mapping and stops working. So there might be a need for mapping the Alt (PC) or Cmd (Mac) key back to Super for this function on Ubuntu. Possibly other DEs. 

Nice thing about Kinto is that it can easily solve a problem David may have when he upgrades to a newer version of Ubuntu: 

```
# Keybindings for gedit (GNOME Text Editor)
define_keymap(re.compile("gedit", re.IGNORECASE),{
        ############################################################
        ###   NEWER VERSIONS of gedit use Ctrl+Alt+PageUp/PageDown (with Alt/Meta key)
        ############################################################
        K("Super-Page_Up"): K("C-M-Page_Up"),             # Go to prior tab/document
        K("Super-Page_Down"): K("C-M-Page_Down"),    # Go to next tab/document
        ############################################################
        ###   OLDER VERSIONS of gedit use Ctrl+PageUp/PageDown (no Alt/Meta key)
        ############################################################
        # K("Super-Page_Up"): K("C-Page_Up"),               # Go to prior tab/document (older versions)
        # K("Super-Page_Down"): K("C-Page_Down"),      # Go to next tab/document (older versions)
},"Gedit - Editor")
```

Lines without Alt key for older versions of gedit, lines with Alt key for newer versions. So he can continue to use the same shortcut he's gotten used to just by switching the lines that are commented out. Don't know exactly where the version cutoff is yet for this shortcut change. Gedit 40 was installed here on KDE Neon when I installed it. 

The descriptions of the shortcuts I've put in the code block above match the key names as reported by `xbindkeys` (Next/Prior). Unfortunately it seems the GNOME apps don't want to respond to the PgUp/PgDn keys on my laptop because they are sending different key codes (KP_Next/KP_Prior) specific to the numpad.

RedBearAK:
@rbreaves 

Do you have a good reference for how to translate `xbindkeys` output into the key names in `key.py`? I googled around for a couple of hours and couldn't find anything useful. I'm trying to remap the numpad PgUp/PgDn keys so that maybe the GNOME apps will respond to them, but I can't figure out what key names to use. Other than KP9/KP3, which doesn't seem to work. `xbindkeys` says they are KP_Next and KP_Prior, but there is nothing like that in `key.py`.

RedBearAK:
This works in Epiphany and Files on eOS, but Code wants the Alt key variant. 

One file manager left to test: Deepin Linux file manager. (dde-file-manager)

All the other file managers work with the shortcut so far, other than SpaceFM which wants the Alt key variant.

rbreaves:
@RedBearAK I don't recall if I ever found a way to directly translate xbindkeys to the names and keycodes in keys.py. I am sure I probably have some documentation in all of the issue tickets that may go into discovering additional keys. Just not something I mess with very much and I really can't recall on top of my head right now.

RedBearAK:
@rbreaves 

Darn. Mainly I'm having trouble figuring out if there is some special way to identify these keys that are only active when Numlock is engaged. I've tried both "Numlock-KP9" and "Shift-KP9", on the theory that it's like calling the "question mark" key by using "Shift+Slash". But `xkeysnail` did not like "Numlock" and the "Shift" variation didn't seem to work. 

If I can't figure this out, I'll always have to use a different computer or and external keyboard to test any shortcuts involving PgUp/PgDn/Home/End, since they're all on the numpad. Seems like any other laptop in this style with those keys on the numpad may also have this problem. Most 15.6" laptops I've looked at for the past couple of years have had numpads just like this one. Wish there were more in that size with MacBook-style symmetrical keyboards. 

The arrow keys on this one are actually dedicated to media functions (Play/Pause/FF/RW) when you hold down the Fn key. I don't know how common that is. But there appears to be no way for this keyboard to send the proper PgUp/PgDn/Home/End key codes. Unless there's some other way more lower-level than `xkeysnail` to change what the numpad keys put out.

RedBearAK:
@dyu @rbreaves 

I just set up an install of Ubuntu 16.04 LTS on the other laptop with the proper PgUp/PgDn keys. It has gedit 3.18.3 installed after doing a full system update. 

The shortcut listed in the menu on this version of gedit also shows that it uses the Ctrl+Alt+PgUp/PgDn shortcuts to move between document tabs, not the Ctrl+PgUp/PgDn shortcuts used in file managers and browsers. 

This would seem to indicate that you may have modified your gedit shortcuts at some point since you installed Ubuntu 16.04. And that there probably hasn't been a change to the gedit shortcuts in any version of gedit, unlike what I was assuming. 

So while you have your own custom mapping that works on your copy of gedit, this should probably the form  for it to take if it were to be added to `kinto.py`, using the Alt key version: 

```
# Keybindings for gedit (GNOME Text Editor)
define_keymap(re.compile("gedit", re.IGNORECASE),{
    K("Super-Page_Up"): K("C-M-Page_Up"),             # Go to prior tab/document
    K("Super-Page_Down"): K("C-M-Page_Down"),    # Go to next tab/document
},"Gedit - Editor")
```

As far as a general addition to file managers, browsers and terminals, I think it may be a good idea. And perhaps also even in the main GUI block. But there is the issue to keep in mind of clobbering the existing Ubuntu workspace navigation shortcuts.

RedBearAK:
@rbreaves 

Alright, so, apparently there simply isn't a way for `xkeysnail` to distinguish between the numpad keys with Numlock disabled and the same numpad keys with Numlock enabled. 

I really can't figure out what I was doing wrong before, but now the KP9/KP3 key names seem to work in `kinto.py` to let me map the numpad KP_PgUp/KP_PgDn keys to PgUp/PgDn shortcuts. It's just that the shortcut does the same thing whether Numlock is on or off. So if you were to try and use the KP9 key in a shortcut all by itself, you would perform the same action when you typed "9" on the numpad with Numlock on and when you typed "PgUp" with Numlock off, because it's the same key. 

I guess the same is true for other keys on the keyboard regardless of Capslock being on or off. It just seems like it should be different somehow. 

Even works in gedit where I was having trouble getting it to work before. And Nautilus. 

I also finally realized something I should have realized days ago. I can use the Fn-Arrow media keys on this keyboard for testing the PgUp/PgDn shortcuts by just using the key names corresponding to those media funtions: 

```
PlayPause = Page_Up
StopCD = Page_Down
PreviousSong = Home
NextSong = End
```

So the fact that my main laptop doesn't have proper PgUp/PgDn keys may no longer be as much of an impediment to testing. 

Oops, maybe spoke too soon. Seems to work well in GUI apps, but the terminals are very inconsistent in how they respond to the mapped numpad keys and media/arrow keys. Some also change based on whether Numlock is on or off, rather than consistently responding to what `xkeysnail` is mapping the shortcut to, like the GUI apps do. 

So, two steps forward, one step back.

RedBearAK:
Fedora 34 with GNOME 40 also uses Super+PgUp/PgDn/Home/End keys for workspace navigation. The only difference from Ubuntu is that the workspaces in GNOME 40 are now oriented horizontally rather than vertically. 

Interestingly, this will absolutely not work on my keyboard without the Kinto remapping enabled. It appears that GNOME shell and GNOME apps simply don't want to recognize the keypad PgUp/PgDn keys. But with the Kinto remapping I can use both the keypad and the arrow keys, for both tab navigation and workspace navigation. 

But once again the terminal in GNOME does not want to cooperate with these keys. This only works in the "GUI" apps.

RedBearAK:
@rbreaves 

OK, I just taught myself, hopefully for the last time, that there is no "Super/Win" key modifier on the keyboard when in the terminal mapping. Thus explaining why the terminals were not cooperating with these shortcuts. 

I've never understood exactly why "Super" is left out of the keyboard modifiers for terminals, and still don't, but at least I understand why Super-key based shortcuts will never work on the origin side of the mapping. Finally. 

So replacing "Super" with "LC" on the left side of the mapping function makes everything work perfectly in terminals and behave just like the GUI apps. Obviously. 

```
define_keymap(re.compile(termStr, re.IGNORECASE),{
    ################################################################
    ###  Keyboards with "real" PgUp/PgDn keys
    K("LC-Page_Up"): K("C-Page_Up"),         # Go to prior tab/document
    K("LC-Page_Down"): K("C-Page_Down"),     # Go to next tab/document
    ################################################################
    ###  Keyboards with numeric keypad (numpad) PgUp/PgDn keys
    K("LC-KP9"): K("C-Page_Up"),             # Go to prior tab/document (numpad Page Up key)
    K("LC-KP3"): K("C-Page_Down"),           # Go to next tab/document (numpad Page Down key)
    ################################################################
    ###  Keyboards with media functions on the arrow/cursor keys
    K("LC-PlayPause"): K("C-Page_Up"),       # Go to prior tab/document (Play/Pause media key on Up Arrow)
    K("LC-StopCD"): K("C-Page_Down"),        # Go to next tab/document (StopCD media key on Down Arrow)
    #
... (snip)
}, "terminals")
```

I was a bit confused because it seemed like the Super key shortcuts were "working" in Konsole, but that's because Konsole and Dolphin have pre-existing tab navigation shortcuts for Ctrl+PgUp/PgDn. So that was an illusion. 

I think I'm getting close to making this work almost everywhere. I'm in KDE Neon at the moment, but with some GNOME apps mixed in, including `gnome-terminal`, and everything is now cooperating as far as I can tell. Moving on to testing in some GNOME environments.

rbreaves:
@RedBearAK Comes down to expected defaults and what I would prefer to happen if a remap isn't accounted for. Do I want to potentially trigger a Super key for something I haven't fully mapped to the Cmd position while using a terminal or do I want it to fallback to the Ctrl key? I choose a fall back to Right Ctrl instead of a fallback to Super is all that is.

Also since there's a difference between the actual Left Ctrl vs the mapped Left Cmd (being Right Control) I can easily add shift on just the Right Control combos. It's all about thinking about the interactions in layers, it does make sense and I am still able to map to some and potentially all expected system level remaps but yes - you have to be aware of the shift in the remaps happening to be successful at it. Putting the Super key though in place of that Cmd position during terminal usage would be a bag of hurt though if you missed a Super-Shift combo or whatever else imo. Eliminating it simplified things - and definitely simplifies things on the Windows side as well where it tries to treat the Super key much more poorly.

RedBearAK:
@rbreaves 

I see. Interesting. You really had to think through a lot of different things when you were creating your mapping scheme. 

And yes, it is all very confusing without some sort of visual reference on which to base a cognitive model of the layering that's going on. I'm picking it apart but it's still like wading through molasses most of the time.

RedBearAK:
Well, I don't know if anyone over there will actually see this as a problem worth spending time on, but I submitted an issue in the GTK Gitlab about the fact that GNOME/GTK applications in my testing are refusing to respect the numeric keypad PgUp/PgDn/Home/End keys on my keyboard, unlike the Qt or other types of applications. 

I saw some other issues that seemed directly related but they have not been getting any attention or even acknowledgement. And the issue in general seems to have been around for quite a long time. So I'll be on maximum patience mode for this one. We will see. 

https://gitlab.gnome.org/GNOME/gtk/-/issues/3868

rbreaves:
> Well, I don't know if anyone over there will actually see this as a problem worth spending time on, but I submitted an issue in the GTK Gitlab about the fact that GNOME/GTK applications in my testing are refusing to respect the numeric keypad PgUp/PgDn/Home/End keys on my keyboard, unlike the Qt or other types of applications.
> 
> I saw some other issues that seemed directly related but they have not been getting any attention or even acknowledgement. And the issue in general seems to have been around for quite a long time. So I'll be on maximum patience mode for this one. We will see.
> 
> https://gitlab.gnome.org/GNOME/gtk/-/issues/3868

I've actually noticed that too.. I have a cheap chinese laptop that uses numpad enter on the regular enter key and some terminals just completely ignore it lol. Obviously I can remap that easily enough.. but still annoying.

RedBearAK:
> I have a cheap chinese laptop that uses numpad enter on the regular enter key and some terminals just completely ignore it

Ew. 

Also, which terminals, if you can recall? I'm curious if they're all GTK-based and therefore explained by this bug.

rbreaves:
In my case it was xfce4-terminal, not sure which framework it uses.

RedBearAK:
Well, Wikipedia says Xfce is based on GTK, so I assume that would include the terminal app. Maybe we have a chance of actually fixing that behavior. 

It might be because I posted a question in the GNOME Reddit, but I actually got a response from a maintainer on that bug report already. I added a bunch of additional information and my observation of how the keypad keys behave in a text field (spoiler alert: very badly). 

Keep your fingers crossed.

RedBearAK:
@dyu 

Gedit and other apps like elementary OS "Code" (probably a fork of gedit), use the Ctrl+Alt variation for tab navigation because the Ctrl key shortcuts perform "text navigation" functions in those types of apps. Surprisingly, these work even with my numeric keypad keys, which refuse to work with the tab navigation shortcuts. 

I only noticed these functions after trying a document with some real text "paragraphs" (multiple lines without a line break character). 

```
Ctrl+PgUp = Go to beginning of first line of paragraph. 
(Won't go past beginning of first line of paragraph, up even if repeated.) 

Ctrl+PgDn = Go to end of last line of paragraph. 
(Won't go past the end of last line of paragraph, even if repeated.) 

Ctrl+Home = Go to beginning of first line of document. 

Ctrl+End = Go to end of last line of document. 
```

Also the Shift+Ctrl variations, and so on. 

Now I have to add these to my mappings for fixing the numeric keypad behavior in GTK apps.

RedBearAK:
@rbreaves 

Since this PR only seems applicable to the submitter's system (his `gedit` shortcuts appear to be different from defaults on any system I've tested), I'd have to recommend this PR be closed without merging. The user will just have to keep their customized block for `gedit` to keep those shortcuts working. 

Meanwhile, PR #499 has lines that should actually cover what this user was trying to do, but works with the default shortcuts of each type of app. So for tab navigation, apps like Nautilus, terminals or web browsers will respond to pysical Ctrl+PgUp/PgDn, and tabbed interface text editors like `gedit` will respond to physical Ctrl+Alt+PgUp/PgDn.

PR #479, opened 2021-04-13
Add GNOME Boxes WM_CLASS for Linux, and a comma

RedBearAK:
Note: I'm trying to consistently leave a comma after the last item in each vertically formatted python list, so that when someone with less python knowledge wants to add something to the list and does it the "lazy way" by just copying the last line, pasting it on the end and editing it to add a new item, they won't cause the "missing comma" python error too easily. Having a comma on the very end of the list with no element after it doesn't seem to cause any harm or python errors. And you can copy/paste the last item as many times as you want without getting into trouble.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -41,7 +41,8 @@ remotes = [
     "org.remmina.Remmina",
     "xfreerdp",
     "VirtualBox Machine",
-    "VirtualBox"
+    "VirtualBox",
+    "Gnome-boxes",
 ]
 remotes = [client.casefold() for client in remotes]
 


rbreaves:
I kinda like putting commas at the beginning to be honest, but python won't break with an extra comma and for most people that looks better to them. Javascript though I think would break with an extra comma.

RedBearAK:
Didn't think of doing it that way. Interesting. But yes, unusual.

rbreaves:
Interesting that I have never heard of this app, but I guess it is linux specific.

RedBearAK:
Yeah, I'm not that familiar with all of the different VM back ends, but I think GNOME Boxes is mainly a simple front end for QEMU. I got tired of Virtualbox one day, it's so complicated to set up a new VM and a number of distros don't work well in it. So I tried Boxes out and it is r-i-d-i-c-u-l-o-u-s-l-y easy to set up a new VM. As in 3 or 4 clicks and it usually works right away. And performance is usually very good. 

I'm messing with Ubuntu Budgie 20.10 in a Boxes VM right now, working on the PgUp/PgDn remapping stuff. Finally figured out how to mostly Macify it smoothly. Nice so far.

rbreaves:
Well.. kairos is functional now btw if you’ve not tried it lol..

https://github.com/rbreaves/kairos

PR #487, opened 2021-04-27
Ctrl(Super)+PgUp/PgDn instead of braces for tab navigation in browsers

RedBearAK:
Firefox in macOS uses Ctrl+PgUp/PgDn to navigate tabs, while Cmd+Braces is for back/forward page navigation. Since the main goal of Kinto is to emulate macOS shortcuts wherever practical, using the Ctrl+Brace shortcuts for tab navigation should be disabled by default. 

The equivalent for tab navigation would then be Super+PgUp/PgDn, since the physical Ctrl key is Super. 

Proof of concept addition of a tweak item to enable/disable the braces for tab navigation will follow shortly.

Tweak item proof-of-concept in PR #489.


RedBearAK:
Closing this because it is superseded by PR #536.

PR #488, opened 2021-04-27
Add tweak item to enable/disable braces for tab navigation in browsers

RedBearAK:
Proof of concept tweak item for Kinto GUI to enable/disable the "BracesNav" lines from PR #487.

It has unnecessary code because it was copy/pasted from another item in the Tweaks dialog, but it demonstrates the intention.


RedBearAK:
Merged into PR #489. 

Closing.

PR #489, opened 2021-04-27
Tweak item to enable/disable braces for tab navigation in browsers

RedBearAK:
Proof of concept tweak item for Kinto Tray to enable/disable the "BracesNav" lines from PR #487.

It has unnecessary code because it was copy/pasted from another item in the Tweaks dialog, but it demonstrates the intention.

### DIFF
--- a/linux/gui/kinto-gui.py
+++ b/linux/gui/kinto-gui.py
@@ -710,7 +710,7 @@ class MyWindow(Gtk.Window):
         restartsvc = False
         self.rightmod =  Gtk.CheckButton('AltGr on Right Cmd')
         self.vsc2st3 = Gtk.CheckButton('ST3 hotkeys for VS Code')
-        self.bracesnav = Gtk.CheckButton('Use braces for tab navigation in browsers')
+        self.bracesnav = Gtk.CheckButton('Use Cmd+Braces for tab navigation in browsers')
         self.caps2esc = Gtk.CheckButton('Capslock is Escape when tapped, Cmd when held')
         self.caps2cmd = Gtk.CheckButton('Capslock is Cmd')
         
--- a/linux/gui/kinto-gui.py
+++ b/linux/gui/kinto-gui.py
@@ -677,7 +677,7 @@ class MyWindow(Gtk.Window):
         win.set_default_icon_list([pixbuf])
 
         win.set_title("Kinto Tweaks")
-        win.set_default_size(350, 200)
+        win.set_default_size(500, 200)
         win.set_position(Gtk.WindowPosition.CENTER)
 
         # Check AltGr - commented out is enabled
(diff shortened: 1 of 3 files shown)


rbreaves:
I'd prefer one of these to be closed and to see both #488 and #489 as a single PR. I don't want to be testing these individually since it is both the same feature.

RedBearAK:
@rbreaves 

Used VS Code to merge patch-18 and patch-19 into patch-20 (PR #489). 

Closing PR #488. 

Leaving #487 so you can optionally accept at least that part for now even if you don't want to merge the new tweak item. I made a mistake on the description though, macOS Firefox uses Cmd+Braces for back/forward, not Ctrl+Braces. But with the Kinto remapping we're talking about the same physical location on the keyboard since PR #440 added shortcuts for Ctrl+Braces and Control_R is on the "Cmd" Alt key.  

The tweak item shouldn't be accepted the way it is, honestly, because I just copied all of the elements of Caps2Esc and got it to the point where it works fine, but I wasn't able to fully parse through the perl line yet to remove the unnecessary bits and change the substitution section to match. 

On the other hand, it doesn't seem to produce any harm in its current state, it's just looking for things that don't exist. The tweak item shows up as enabled if the lines are uncommented, and vice versa, and it reverses the state every time I toggle it and close the tweaks dialog. 

Would be good to have an "Apply" button or something in the tweaks dialog. It's a little unintuitive that the change is only applied when you close the dialog window. That doesn't match either the KDE/Windows paradigm of always having to click "Apply" or "OK", or the GNOME/macOS paradigm of having options instantly change. Since the tweaks require restarting the `xkeysnail` service in the background, I think it makes the most sense to have an "Apply" button.

rbreaves:
Well I am also wondering why it is important to toggle it - if it is the default under macOS then I feel like it should just be mapped then.

RedBearAK:
> Well I am also wondering why it is important to toggle it - if it is the default under macOS then I feel like it should just be mapped then.

Well. The user who pushed PR #440 seemed really happy to have it added. And I'm sure, like he said, that he isn't the only person to use Cmd+Braces for tab navigation. It is sort of an intuitive shortcut for tabs. 

So I wanted to make sure those users had the option to simply turn it back on if they want it, before I asked for it to be disabled by default. 

That's all.

rbreaves:
I see, you want it disabled by default and some users want it enabled by default. Ok, well I will try and get this tested in the next day or 2 and will likely merge it then. I may enable it by default though - anything that aligns with how things are on a mac gets priority as far as I am concerned otherwise it leaves users guessing on what behaviors will be here by default and what won't.

RedBearAK:
@rbreaves 

There may be some confusion here. 

PR #440 is kind of strange because it maps the physical equivalent of Cmd+Braces to (logical) Ctrl+PgUp/PgDn. That shortcut on the right side of the function (Ctrl+PgUp/PgDn) is the same as the actual default physical shortcut on Mac for moving between tabs. So, if I read it correctly, they are saying they somehow changed the shortcut for tab navigation on their browser in macOS to actually be Cmd+Braces. Which in Firefox seems to require a 3rd party browser extension. 

But since Cmd+Braces is page navigation by default on a Mac, the mapping of Braces to PgUp/PgDn should be disabled by default. 

The equivalent of Ctrl+PgUp/PgDn on a Mac (for tab navigation) is Super+PgUp/PgDn, which should be enabled by default. This is included here, but should be harmlessly superseded by an addition to the General GUI block that I'm planning to put in another PR. 

I guess the Tweak item is really optional, since the user could just modify their copy of Firefox to change the shortcut the same way they did in macOS.

RedBearAK:
Closing this because the tweak option really doesn't need to exist. The correct shortcuts for tab nav in web browsers and the Finder in macOS is Cmd+Shift+Braces and that will be resolved by PR #536 and PR #541, while leaving the Cmd+Braces option available but disabled in `kinto.py` for those who want to re-enable it.

PR #498, opened 2021-05-04
Add support for sysvinit & mxlinux w/ xfce

rbreaves:


rbreaves:
This just keeps growing.. 50 lines on that last commit, but it did fix like 3 or 4 things, including a bug where some children processes weren't being ended properly and could accumulate - particularly on consecutive installs as what tends to happen during dev'ing or adding shortcut keys.

Issue #486, opened 2021-04-24
Fixes for numeric keypad PgUp/PgDn/Home/End and media key arrows


RedBearAK:
I don't think this is quite done, and I'm still wondering if there isn't a smarter or more compact way to do this, but I'm posting this as the beginning of a proposed solution to the problem of my numeric keypad navigation key codes (numpad arrows and PgUp/PgDn/Home/End key functions that are active when Numlock is off) not being treated like real navigation/cursor keys. It's a little tricky even if there were a generalized way to do it since some of the lines are remapping Ctrl to Super and vice versa. 

Numpad keys by themselves are left out since that would interfere with actually using the numpad for numbers. Might add it later as an option, like with the media function arrow keys. Some people just may not care to ever use the numpad for numbers and want it to always behave like navigation/cursor keys. 

Includes work on being able to use arrow keys like those on my Acer Aspire A515-43, which are dedicated to the Play/Stop/FF/RW media functions, as if they were also real PgUp/PgDn/Home/End keys when used with any combination of modifiers (Shift/Ctrl/Alt/Super). Well, any one or two modifiers. Haven't done combinations of three or four modifiers. Not sure if I want to go that far. 

This is still very much a work in progress. Might just keep modifying this first post with updated code rather than adding new posts, except to answer questions if there are any. 

For the media keys fix, the lines for the "base key" (the key without any modifiers besides "Fn" key that activates the alternate hardware function) are commented out so the media keys are still usable by themselves for their intended media functions. But those lines could optionally be enabled if the user doesn't care about the media functions and really wants to use the arrows as PgUp/PgDn/Home/End in general use rather than as media keys. They are labeled at the end: [ `(fix_media_keys_on_arrows_base_key)` ]

_It has to be noted here that these "base key" lines, if enabled, would naturally interfere with any keyboard that has dedicated media keys. I've seen some laptop keyboards that put the media keys above the rest of the numpad keys. It would turn those dedicated media keys into PgUp/PgDn/Home/End. So it's only useful for an odd keyboard like mine and should otherwise remain disabled._

The rest of the media key lines are for when the media arrow keys are used with other modifiers. They stop being "media keys" and get mapped to PgUp/PgDn/Home/End. They have a keyword at the end of each line that could be used to disable or enable all lines with `sed` in a Kinto tweak option. [ `(fix_media_keys_on_arrows)` ]

For use in the "General GUI" code block: 

```
###########################################################################################################
[...]

PR #466, opened 2021-04-06
Fix go to line, recent files popup IntelliJ shortcuts

OlegSmelov:
| Shortcut           | macOS | Linux/Windows |                                       |
|--------------------|-------|---------------|---------------------------------------|
| Go to line         | ⌘ L   | Ctrl + G      | Fixed                                 |
| Recent files popup | ⌘ E   | Ctrl + E      | Follows the general pattern - removed |

I used [this PDF](https://resources.jetbrains.com/storage/products/intellij-idea/docs/IntelliJIDEA_ReferenceCard.pdf) for reference.

PR #499, opened 2021-05-05
Fixes for numeric keypad nav/cursor keys and media key arrows

RedBearAK:
See issue #486 for additional info/discussion.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -503,6 +503,262 @@ define_keymap(lambda wm_class: wm_class.casefold() not in remotes,{
     K("Alt-Delete"): K("C-Delete"),               # Delete Right Word of Cursor
     # K(""): pass_through_key,                      # cancel
     # K(""): K(""),                                 #
+
+    ###########################################################################################################
+    ###  Fixes for media functions on arrow/cursor keys in GTK apps 
+    ###  Uncomment first 4 lines to make media key arrows always act as PgUp/PgDn/Home/End keys. 
+    ###  WARNING: Base media key remapping this way will interfere with any 
+    ###           dedicated "media" keys that use the same key codes! 
+    ###========================================================================================================
+
+    ###  Media keys with no modifier keys (base key)
+    # K("PlayPause"):       K("Page_Up"),   # PlayPause media function (fix_media_keys_on_arrows_base_key)
+    # K("StopCD"):          K("Page_Down"), # StopCD media function (fix_media_keys_on_arrows_base_key)
+    # K("PreviousSong"):    K("Home"),      # PreviousSong media function (fix_media_keys_on_arrows_base_key)
+    # K("NextSong"):        K("End"),       # NextSong media function (fix_media_keys_on_arrows_base_key)
+
+    ###  Media keys with 1 modifier key
+    K("Super-PlayPause"):               K("C-Page_Up"),                 # (fix_media_keys_on_arrows)
+    K("Super-StopCD"):                  K("C-Page_Down"),               # (fix_media_keys_on_arrows)
+    K("Super-PreviousSong"):            K("C-Home"),                    # (fix_media_keys_on_arrows)
+    K("Super-NextSong"):                K("C-End"),                     # (fix_media_keys_on_arrows)
+
+    K("Shift-PlayPause"):               K("Shift-Page_Up"),             # (fix_media_keys_on_arrows)
+    K("Shift-StopCD"):                  K("Shift-Page_Down"),           # (fix_media_keys_on_arrows)
+    K("Shift-PreviousSong"):            K("Shift-Home"),                # (fix_media_keys_on_arrows)
+    K("Shift-NextSong"):                K("Shift-End"),                 # (fix_media_keys_on_arrows)
+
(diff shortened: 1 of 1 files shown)


rbreaves:
Will be testing this shortly. Will likely merge this before the merged commit that includes this and several other fixes you've been working on. May let it sit in master for a day or so before merging the rest. This is without a doubt the largest PR I have had to date lol.

RedBearAK:
@rbreaves 

Definitely recommend testing the individual PRs rather than looking at the merged version. The merged PR only exists to make it easier for me to get the `kinto.py` that I want from a central location when installing on new systems. 

Half of this PR basically only applies to all the GTK environments or applications that refuse to acknowledge the numpad nav keys as part of shortcut combos, so that half would normally serve no purpose in Windows. But with the advent of WSL2 and GUI Linux apps, that may not be true for long. 

To test its behavior all you need is some GTK apps and any keyboard with a numeric keypad. Anything with tabs makes it easy to see the difference. 

The other half applies to the atypical Acer arrow keys that have media functions rather than the usual PgUp/PgDn/Home/End functions most laptop keyboards would have on those keys. In short, it makes keyboard shortcuts using my arrow keys behave as you would expect on an Apple laptop keyboard using the same keys. 

Since the basis of the translation is the "media" key names rather than the Up/Down/Left/Right arrow key names, and it's extremely unlikely anyone with a different type of keyboard would attempt to use independent media keys in normal places (like the top keyboard row) as part of keyboard shortcuts, I don't think those remappings can cause any harm on non-Acer type keyboard layouts. But, since it is specific to a type of keyboard rather than any particular application, that part of the PR should be equally applicable to Windows. 

To test this part on a non-Acer style keyboard layout you would need to attempt to use the "media" keys on your keyboard as part of some keyboard shortcut combos, as if they were PgUp/PgDn/Home/End keys. Good luck with that. 

I'm going to be setting up a Win10 VM shortly and I'll see if I can finally get familiar enough with the AutoHotkey format to start translating this PR for Windows. If you're concerned about there being a mismatch between the Linux and Windows configs, just don't merge this yet. I can always use my own merged PR for now. 

The main PR that really needs your attention if you have some time is the Fedora install fix (#446) from @prosac. It's simple and worked for me now on several Fedora installs, but it needs to be rebased on the latest Kinto installer code to get rid of the conflicts.

rbreaves:
Yea, and I was just trying to see what behaviors exist on macOS and make sure things align, but it sounds like this PR resolves some specific hardware and GTK issues rather than more typical remaps that can be universally applied - although like you say may not cause any issues on different layouts but still. 

I also have a few commented lines in my Windows kinto.py that are laptop specific. A better way of detecting and applying laptop or keyboard specific keymaps would be good.

rbreaves:
#466 is now merged, there were some additional changes that I had to make to ensure compatibility that went beyond the conflicts. Apparently symlinks would fail under SELinux, at least under Fedora 34, so I just copy the service file directly into place and it works fine that way.

RedBearAK:
> #466 is now merged, there were some additional changes that I had to make to ensure compatibility that went beyond the conflicts. Apparently symlinks would fail under SELinux, at least under Fedora 34, so I just copy the service file directly into place and it works fine that way.

Outstanding. Thanks for that. Although I think you meant #446 not #466, that's about IntelliJ. 

Now I won't have to go track down that branch every time I set up a new system with Fedora 34. Which has been happening a lot lately. 

Happy 4th!

rbreaves:
Thanks! And happy 4th to you as well & yea I meant 446 lol.  I guess he must have put that in at a time when I had a lot going on 😅. Trying to slow down & catch up but it’s hard atm.

joshgoebel:
You should perhaps open an issue on my repo to discussion why this insane number of combos is needed and what might be done about that.

rbreaves:
I need to test this soon, even if it changes in the switch to keyszer to something simpler I assume this will still be pretty valid.

RedBearAK:
@rbreaves 

> I need to test this soon, even if it changes in the switch to keyszer to something simpler I assume this will still be pretty valid.

I still need to get an answer from Josh if he can think of a way to limit a `modmap` to operating only on a set of keys when they are combined with at least one modifier, and NOT on the base key. I don't know if a `when` test on the `modmap` can manage that. If it can't, then a full static mapping like this is necessary for the Numpad fix. 

In this way, the Numpad fix is unlike the "Media Arrows Fix" that I was able to reduce to a simple `modmap` that will only be activated if the user chooses to do so. 

FYI: 

You will need to make sure you use a GTK3 app to test. The issue with Numpad nav keys getting ignored by GTK apps seems to be fixed with any app based on GTK4. 

And I always get a little confused on this point but I'm pretty sure that the issue only appears when you attempt to combine a modifier with the Numpad nav keys, like Shift+PgUp. The base keys should actually work, even in GTK3 apps.

joshgoebel:
> if he can think of a way to limit a modmap to operating only on a set of keys when they are combined with at least one modifier, and NOT on the base key.

That sounds weird at first blush and something better done just by building the proper map tables with a loop... I'm not sure modmaps should ever be conditional based on key state (lock state perhaps, keystate, very unsure)...

PR #506, opened 2021-05-09
Quit Chromes with Cmd+Q, fix Chromes preferences shortcut

RedBearAK:
Chromes on Linux refuse to respond to Cmd+Q (physical Alt+Q, logical RC+Q under Kinto remapping), so it needs to be mapped to Alt+F4 (M-F4). 

To get Chrome browsers to respond properly to Cmd+Q (Quit), I had to rearrange the order of the `chromeStr` block so that it can take precedence over the more "general" browser block which defines `K("RC-Q"): K("RC-Q"),` and was therefore interfering. 

The preferences shortcut for Chromes was not working, opens the Edit menu but can't proceed further because the "Preferences" item doesn't seem to have a shortcut letter anymore. (Moving through the Edit menu with arrow keys shows no underlined letter in "Preferences" and there is no shortcut visible.) So I implemented a multi-keystroke macro solution modeled on the existing method for opening Firefox preferences. It opens a new tab and just types "chrome://settings" and Enter key. Seems to work just as reliably as the equivalent Firefox preferences shortcut. Possibly better, because it's a bit shorter (enters quicker). Hard to tell. 

Reworked some of the related comments and code block labels based on the Finder mod comments, for clarification of the purpose of each code block in the "browsers" section.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -388,7 +388,27 @@ define_keymap(re.compile(filemanagerStr, re.IGNORECASE),{
 ### END OF FILE MANAGER GROUP OF KEYMAPS ###
 ############################################
 
-# Keybindings for Browsers
+# Keybindings overrides for Chromes/Chromium browsers 
+# (overrides some shortcuts from General Web Browsers block below)
+define_keymap(re.compile(chromeStr, re.IGNORECASE),{
+    K("RC-comma"): [
+            K("C-t"),K("c"),K("h"),K("r"),K("o"),K("m"),K("e"),K("Shift-Semicolon"),K("Slash"),K("Slash"),
+            K("s"),K("e"),K("t"),K("t"),K("i"),K("n"),K("g"),K("s"),K("Enter")],
+    K("RC-q"): K("M-F4"),
+}, "Overrides for Chrome-based Browsers")
+# Opera C-F12
+
+# Keybindings overrides for Chromes/Chromium browsers 
+# (overrides some shortcuts from General Web Browsers block below)
+define_keymap(re.compile("Firefox", re.IGNORECASE),{
+    K("C-comma"): [
+        K("C-T"),K("a"),K("b"),K("o"),K("u"),K("t"),
+        K("Shift-SEMICOLON"),K("p"),K("r"),K("e"),K("f"),
+        K("e"),K("r"),K("e"),K("n"),K("c"),K("e"),K("s"),K("Enter")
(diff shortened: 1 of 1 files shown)


RedBearAK:
@rbreaves 

Chrome code block must be moved before the "general browsers" code block in order to allow overriding individual shortcuts that exist in the "general browsers" code block but aren't working properly in Chrome(s).

rbreaves:
I believe this was part of the merged PR PR which I merged earlier, so I assume it is in master and is why I closed this one.

RedBearAK:
> I believe this was part of the merged PR #537 which I merged earlier, so I assume it is in master and is why I closed this one.

AFIK, the only merge I've seen is #594, the updates to the Finder Mods. There were no conflicts there, and it had nothing to do with the web browser blocks. 

I closed 537, since it's just really just meant for my own use. 

#506 has always had a conflict even before the merge of 594, because of the reorganization of the order of the browser code blocks. You'll have to re-open and merge #506 and allow the changes to resolve this. I can't resolve it from my side, as it is a fundamental change in the web browser group of code blocks. 

It just takes Chromes and Firefox blocks and puts them before the "general" block.

RedBearAK:
@rbreaves 

Current master of this file has web browser blocks in this order (which causes issues): 

General
Firefox
Chrome(s)

PR #506 has them in this order: 

Chrome(s)
Firefox
General

It has to be in this order for the same reason individual GUI apps need to have their overrides come before the "General GUI" block. Or oddball terminals need to have their overrides come before the main `termStr` block.

rbreaves:
This reorder has been done, although the order is now firefox, chrome and then general. I understand the merged PR was more for your usage than to be merged all at once. I felt it was ok to merge at once & I think the assumption that you merged the actual commit that made up 506 was proper (aka used git the normal merge, new branch then rebase or cherry-pick method).

I'd suggest merging the actual branch or cherry picking it next time versus trying to re-make these changes by hand as these sort of inconsistencies, in my experience, only occur when one tries to re-duplicate their work by hand versus merging branches or using git's cherry-pick feature.

I had to force myself in the past to use git this way because I also initially resisted using git's features until I realized that the git revert feature worked better than me trying to roll things back by memories or by evaluating things manually and backing things out by hand. Things often got missed, big or small, so after that I made sure that I made use of these features in git, along with forcing team members to slow down with me so that we could move faster in the future with more consistency.

PR #526, opened 2021-06-04
Fix zoom in/out (font size) shortcuts in terminals

RedBearAK:
On macOS the terminal, like a web browser, lets you "zoom" in and out with Cmd+Equal and Cmd+Minus. It doesn't force you to type Cmd+Shift+Equal to get to the "Plus" keycode on the Equal key. On Linux of course, terminals want you to use Ctrl+Shift+Equal (Plus) to zoom in (increase font size) and then Ctrl+Minus without the Shift key to decrease the font size (zoom out). This is very awkward. 

Mapping RC-Minus to C-Minus in the termStr block instead of to C-Shift-Minus lets the user zoom in and out in terminals by just pressing the equivalent of Cmd-(Equal/Minus) keys, without ever worrying about the Shift key. Same as terminals in macOS, and GUI web browsers. 

Since this is probably just one of the keys that was blanket remapped from RC to C-Shift, I don't know of any reason this should be harmful.


rbreaves:
I have been noticing this too lately, I guess I have been lazy. Thanks for catching it as well and getting it added!

PR #538, opened 2021-06-28
Add Cmd+K to clear both terminal and chrome Devtools console in linux

rajibds:
Shortcut | macOS | Linux |  
-- | -- | -- | --
Clear Terminal | ⌘ K |  | 
Clear Chrome DevTools Console | ⌘ K | Ctrl + L |

<br class="Apple-interchange-newline">


rbreaves:
Withholding this one for now - the Firefox remap looks complicated and I would want this worked out for both Windows and Linux at the same time so that parity is maintained.

PR #551, opened 2021-07-27
bugfix: inputing semicolon on alt+up/down

HADMARINE:
when ALT  + UpArrow or ALT + DownArrow
semicolon is inputed after PgUp / PgDown


HADMARINE:
#552 will cover this PR

Issue #535, opened 2021-06-27
Chrome tab switching with cmd-option-left/right


Grsz:
**Is your feature request related to a problem? Please describe.**
I tried to set the following in `kinto.py`:
`470    K("Super-Tab"): K("LC-M-Right"),                  # Default not-chromebook`

**Describe the solution you'd like**
I'm expecting pressing the two keys next to space to the left if pressed together with a horizontal arrow key result in moving to the tab in direction corresponding to the selected one like in mac.

RedBearAK:
@Grsz @rbreaves 

I verified that the Cmd+Option+Left/Right shortcut works on macOS, in Opera, Firefox, and Google Chrome. 

So how would we do this in `kinto.py`? 

The keys on the left side of the function get mapped to the keys on the right side. To use Left/Right you'll need two lines. 

The currently working shortcut for browser tab navigation on Linux is Ctrl+PgUp/PgDn (when keyboard remapping is disabled). So that has to be on the right side to make it work. (Ctrl+PgUp/PgDn is also a working shortcut for tab navigation on macOS.)

Something like this should do it: 

```
    K("RC-M-Left"):     K("C-Page_Up"),         # Go to prior tab
    K("RC-M-Right"):    K("C-Page_Down"),       # Go to next tab
```

This code block is working for me on Fedora 34 (GNOME 40) with Firefox and Chromium.
```
# Keybindings for General Web Browsers
define_keymap(re.compile(browserStr, re.IGNORECASE),{
    K("RC-Q"): K("RC-Q"),           # Close all browsers Instances
    K("M-RC-I"): K("RC-Shift-I"),   # Dev tools
    K("M-RC-J"): K("RC-Shift-J"),   # Dev tools
    K("RC-Key_1"): K("M-Key_1"),    # Jump to Tab #1-#8
    K("RC-Key_2"): K("M-Key_2"),
    K("RC-Key_3"): K("M-Key_3"),
    K("RC-Key_4"): K("M-Key_4"),
    K("RC-Key_5"): K("M-Key_5"),
    K("RC-Key_6"): K("M-Key_6"),
    K("RC-Key_7"): K("M-Key_7"),
    K("RC-Key_8"): K("M-Key_8"),
    K("RC-Key_9"): K("M-Key_9"),    # Jump to last tab
    K("RC-M-Left"):     K("C-Page_Up"),         # Go to prior tab
    K("RC-M-Right"):    K("C-Page_Down"),       # Go to next tab
[...]

PR #568, opened 2021-09-12
Add support for Firefox Developer Edition

lydell:
I made this change `~/.config/kinto/kinto.py` and restarted the service. Then cmd-option-left/right worked in Firefox Developer Edition as well – https://github.com/rbreaves/kinto/issues/535#issuecomment-917612176

I’m quite new to Kinto so I’m not 100% sure this is the correct fix, but it seems to work.

Thanks for Kinto btw! It’s awesome.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -64,6 +64,7 @@ browsers = [
     "Discord",
     "Epiphany",
     "Firefox",
+    "Firefox Developer Edition",
     "Google-chrome",
     "microsoft-edge",
     "microsoft-edge-dev",


RedBearAK:
@lydell 

LOL. I didn't notice your PR before writing up a reply in the other thread for how to do this. Saved me the trouble of putting together the same PR myself. 

He is usually pretty quick to merge simple PRs like this that just add a new WM_CLASS name to a group.

rbreaves:
Thanks @lydell , and indeed - changes that just add additional application names to the list are approved quickly as no real testing needs to occur.

PR #593, opened 2021-12-21
Shortcut for LibreOffice Find and Replace dialog

RedBearAK:
Shortcut for the Find & Replace dialog in both LibreOffice and Microsoft Excel in Windows or Linux is Ctrl+H. This conflicts with the remapped Cmd+H (Ctrl+H) shortcut used by macOS for Hide Window. 

Excel on macOS resolves the conflict with Cmd+H by using Cmd+Shift+H as an alternative. 

LibreOffice on macOS resolves the conflict with Cmd+H by using Cmd+Option+F, unlike Excel. (Ctrl+Alt+F on a Kinto remapped PC keyboard.) 

The Ctrl+Alt+F shortcut is useful as a system shortcut to open a file browser window, as a match for the common system shortcut of Ctrl+Alt+T to open a terminal window. I am therefore leaving that line commented out by default here, and choosing instead to enable the shortcut that matches Excel, which more users coming from macOS to Linux will probably be familiar with. 

So the remapping that will be enabled by default in all LibreOffice apps with this change is the equivalent of Cmd+Shift+H (Ctrl+Shift+H) to Ctrl+H to activate the Find & Replace dialog.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -449,6 +449,17 @@ define_keymap(re.compile(chromeStr, re.IGNORECASE),{
 }, "Browsers")
 # Opera C-F12
 
+define_keymap(re.compile("libreoffice*", re.IGNORECASE),{
+    # K("RC-M-f"):                K("C-h"),           # Find & Replace dialog (LibreOffice for macOS)
+    # Uncomment the line above to enable the official LibreOffice for macOS shortcut
+    K("RC-Shift-h"):            K("C-h"),           # Find & Replace dialog (Microsoft Excel for macOS)
+    # LibreOffice for macOS uses Cmd+Option+F, Excel for macOS uses Cmd+Shift+H
+    # Alternatives are required to avoid conflicting with Cmd+H (Hide Window)
+    # Ctrl+Alt+F is useful as a system shortcut to open a file browser, 
+    # so Excel's Ctrl+Shift+H is also mapped here as an alternative. 
+    # 
+},"LibreOffice")
+
 define_keymap(re.compile("ulauncher", re.IGNORECASE),{
     K("RC-Key_1"):      K("M-Key_1"),      # Remap Ctrl+[1-9] and Ctrl+[a-z] to Alt+[1-9] and Alt+[a-z]
     K("RC-Key_2"):      K("M-Key_2"),


rbreaves:
Option+cmd+space opens finder on macOS.
www.howtogeek.com/661251/how-to-open-finder-with-a-keyboard-shortcut-on-mac/

I don’t mind keeping multiple shortcuts for somethings that maybbe heavily used.. but I think I’d like to keep the original shortcut from macOS on excel & libreoffice. Things are dynamic so when it’s not in active use Kinto can still open finder in the Linux way too.

Holding off on this merge for now. Would like to test & see what feels right to me.

RedBearAK:
> Option+cmd+space opens finder on macOS. [www.howtogeek.com/661251/how-to-open-finder-with-a-keyboard-shortcut-on-mac/](http://www.howtogeek.com/661251/how-to-open-finder-with-a-keyboard-shortcut-on-mac/)
> 
> Holding off on this merge for now. Would like to test & see what feels right to me.

@rbreaves 

Ah, silly of me as usual to have used macOS since Cheetah and still failed to realize there is a system shortcut to get a Finder window. Or maybe I tried it long ago and didn't like it. When I test this on the Big Sur install that I'm temporarily booting on my 2011 MacBook Pro, it brings up a Finder window but it's always in a search mode, with the search field highlighted and showing no content. This makes sense with the shortcut being just a variation of the Spotlight shortcut. It's "search, but in a Finder window". 

So I would not consider this exactly the same as just bringing up a new Finder window showing whatever the user has configured as the "home" location (I usually make Finder show me the whole computer instead of home, but on Linux this is somewhat nonsensical so I stick with a Nautilus window forced to always show me my actual home folder.) 

> I don’t mind keeping multiple shortcuts for somethings that maybbe heavily used.. but I think I’d like to keep the original shortcut from macOS on excel & libreoffice. Things are dynamic so when it’s not in active use Kinto can still open finder in the Linux way too.
> 

Re-reading this I actually don't know precisely what you mean by this. Sounds like you're fine with leaving both active, instead of just the Excel shortcut? Does that mean you're agreeing with my assessment that, since there is basically no access to "Excel for Linux", many users would have a muscle memory for the Excel shortcut while using LibreOffice Calc? 

Actually I'm pretty sure this F&R shortcut works the same way with all the Office apps, and all the LibreOffice apps, and the remap is for `"libreoffice*"` so it covers Writer and everything. So I should probably not focus so much on just Excel vs. Calc. It's really about functionality in the whole office suite. 

I'm fine with both being active, and maybe even switching to using `Ctrl+Alt+Space` to open a file browser. The `Ctrl+Alt+F` was just set up to match the terminal shortcut so that it was easy to remember, but that's thinking from the Linux perspective. If the goal is to match shortcuts on macOS as perfectly as possible, it makes more sense to apply the `Ctrl+Alt+F` to LibreOffice to match LibreOffice on macOS, with the Excel shortcut just there as a convenience. 

Assuming this is what you meant, I'll uncomment the LibreOffice line, and maybe remove some of the unnecessary comment lines.  

Hope you're getting some good time off for the holidays.

PR #597, opened 2022-01-06
Updates to remaps for Ulauncher

RedBearAK:
Talked to a Ulauncher dev and determined that the Ulauncher search results are limited for the foreseeable future to a max of 17 items. So the screen size is not relevant and there is no reason at present to remap beyond Ctrl+h. So this keymap block can be trimmed down a bit. 

Remapping Ctrl+a and Ctrl+c of course interferes with quickly selecting and copying text within the search field, so it seems wise to disable those by default, and the trimmed list will no longer have Ctrl+v as a remap, so it won't interfere with pasting text into the search field. I don't see any good way around these conflicts, except maybe shifting the whole remap scheme to using Ctrl+Shift instead of just the Ctrl key. But that wouldn't be very intuitive. 

Most desired results will tend to be near the top of the results list anyway, so the primary benefit is still retained by having 1-9 mapped, plus the remaining letters. 

Adding `KEY_0` to the numerical key remaps, which will have no effect for now since Alt+0 doesn't exist, but I'll be asking the Ulauncher devs to add it as a shortcut in the app, so I hope it will be a usable remap in a future Ulauncher release.


rbreaves:
I mean really.. alt hasn't gone anywhere it is just one key over, so I am not really sure why I even accepted any of these ulauncher remaps tbh - it doesn't really line up w/ anything from macOS. Looks like it just fulfills your specific memory memory muscle of using the original Alt keys position than anything else. It's already been merged though and looks like it makes sense to allow for these changes.

RedBearAK:
@rbreaves 

I have to admit that's true, Spotlight doesn't have an equivalent as far as I know. Not sure if something like Alfred or Quicksilver on macOS had a sort of Cmd+1-9 scheme to launch things. I feel like I have used something like that in the past but it's been years since I felt the need for a third-party app launcher on macOS. 

Appreciate you being flexible on this one. I'll try to stay away from anything else that doesn't have an explicit counterpart in macOS.

rbreaves:
I've not used ulauncher any so far but when I do use similar tools it is Albert, which reminds me, I should add Albert to sorun.me.

https://github.com/rbreaves/sorun/blob/dev/configs/ubuntu_budgie.yaml

PR #599, opened 2022-01-09
Hardcoded python3 path, avoids issue w/ homebrew

rbreaves:


rbreaves:
Need to add in a prompt for the user to select which Python3 binary they want to use since some users may use uncommon paths or python inside of some sort of virtual env path.

```
whereis -b python3
python3: /usr/bin/python3 /usr/lib/python3 /etc/python3 /usr/share/python3 /home/linuxbrew/.linuxbrew/bin/python3
```

PR #600, opened 2022-01-12
Cmd+Shift+Braces for tab/view nav in codeStr apps

RedBearAK:
Cmd+Shift+Braces (in addition to the already existing remap of Opt+Cmd+Left/Right) is a verified working alternate shortcut for tab (view) navigation in VSCodium on macOS. 

While Cmd+Braces advances and retracts the indent/tabbing level, these Cmd+Shift+Braces shortcuts (logical Ctrl+Shift+Braces, physical Alt+Shift+Braces on PC keyboard) appear to do nothing in the Linux version of VSCode without this remap. 

Enabling this will bring the Cmd+Shift+Braces tab navigation behavior in the "codes" apps in line with the same behavior in macOS for web browser tabs, terminal tabs and file manager tabs.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -606,10 +606,12 @@ define_keymap(re.compile(codeStr, re.IGNORECASE),{
     K("M-Shift-Left"): [K("M-F19"),K("C-Shift-Left")],      # Select Left of Word
     K("M-Shift-Right"): [K("M-F19"),K("C-Shift-Right")],    # Select Right of Word
 
-    # K("C-PAGE_DOWN"): pass_through_key,         # cancel next_view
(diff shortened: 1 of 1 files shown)


rbreaves:
Reminds me that I still need to fully script a better Sublime Text key remap to vscode solution.. I have a manually method I keep running through but it sure would save me time because what I have in Kinto currently doesn't quite compare to the Sublime Text key remap. It is a very specific thing to me though.. I don't think anyone has ever asked me to have Sublime Text hotkeys in vscode lol.. and the macOS sublime text remaps at that, which are a good bit different from Windows or Linux.

PR #602, opened 2022-01-16
Block some Backspace and Delete shortcuts in file managers

RedBearAK:
Minor update to Finder mods: 

The macOS Finder performs no action and throws a "bonk" error noise when the Backspace or Delete keys are used without Command, or when pressing Cmd+Delete (the PC Delete "forward delete" key, not the Mac Delete "Backspace" key). This update blocks those keys from having any action in file managers. Finder's delete shortcut is only Cmd+Delete (Backspace), which was already enabled. 

Blocking these keys will conform more closely to Finder behavior, and act as a safeguard against accidental navigation or deletions.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -412,6 +412,9 @@ define_keymap(re.compile(filemanagerStr, re.IGNORECASE),{
     K("RC-Super-o"):    K("RC-Shift-o"),        # Open in new window (or tab, depends on FM setup) [not universal]
     K("RC-Backspace"):  K("Delete"),	        # Move to Trash (delete)
     K("RC-D"):          [K("RC-C"),K("RC-V")],  # Mimic Finder's Duplicate command (Copy, then Paste)
+    K("Backspace"):     None,                   # Block Backspace key from performing any action (error in macOS)
+    K("Delete"):        None,                   # Block Delete key from performing any action (error in macOS)
+    K("RC-Delete"):     None,                   # Block Ctrl+Delete from performing any action (error in macOS)
     ###########################################################################################################
     ###  To enable renaming files with the Enter key, uncomment the two keymapping lines just below this.   ###
     ###  Use Ctrl+Shift+Enter to escape or activate text fields such as "[F]ind" and "[L]ocation" fields.   ###


rbreaves:
Hmmm… I mean their laptops don’t have a dedicated delete key. Their full size keyboard I think does. I kinda don’t mind the delete key working even though it might not under macOS 😅.

You’re also not allowed to cut & paste files the same way as Linux or windows.. it’s cmd+c & cmd+alt+v vs Ctrl+x & Ctrl+v.

RedBearAK:
@rbreaves 

The "forward delete" function can be accessed on Mac portables with Fn+Delete. Of course the full size keyboards have it separately, as part of the PgUp/PgDn/Home/End/Del/Ins group. 

But I don't think that's quite the right perspective from which to look at this. On a lot of PC laptop keyboards the Delete key is right above the Backspace key. 

> You’re also not allowed to cut & paste files the same way as Linux or windows.. it’s cmd+c & cmd+alt+v vs Ctrl+x & Ctrl+v.

Yes, you can't "cut" a file in Finder, that's the secret shortcut if you want Finder to "move" the pasted file and delete the original. But that actually is part of the Apple philosophy that led them to make these shortcut decisions long ago. They blocked the Backspace and Delete keys from functioning in Finder without a modifier, and probably just never implemented an action for Cmd+[PC Delete key]. Although, that also throws a "bonk" error and flashes the screen (accessibility option), so maybe they also meant to treat that as an error rather than just having no default action enabled. 

Just like making it non-obvious how to "cut" and paste files rather than more safely copying files, this was obviously meant to help inexperienced computer users avoid making mistakes and accidentally misplacing or deleting files when they might not have meant to. 

You can always comment out the line blocking the delete key action in your own config if you aren't already in the habit of doing Ctrl+Backspace, but I'm really trying to avoid unexpected behaviors in my Finder mods. Especially possibly destructive behaviors like deletions, and suddenly navigating without intending to. I'm kind of embarrassed I didn't notice and block these keys in the original iteration of the Finder mods. 

In Windows it will be part of my AHK-based Finder mods (I have made a bit of progress in this area) to throw the same "bonk/beep" for those keys that macOS does, to conform to this deliberate Finder behavior that pressing those keys alone is a user error. It's nice that AHK has such an option available.

rbreaves:
I actually do like you bringing up the bonk/beep thing now that I think about it.. we can implement that still with linux too..

Just setup a shell file in the ~/.config/kinto directory with `chmod +x ./bonk.sh` and map that as a custom shortcut under Keyboard.

```
#!/bin/bash
aplay /usr/share/sounds/sound-icons/capital
```
Use either these for a beep I guess or not if you find something better.
aplay /usr/share/sounds/sound-icons/beginning-of-line
aplay /usr/share/sounds/sound-icons/empty-text

RedBearAK:
@rbreaves 

Yeah, that's the sort of additional customization that 99% of users would never bother with implementing, if they have to do it themselves. And different distros will have different sound files available, as you point out. On the other hand, if `xkeysnail` could be patched to run a shell file on a shortcut, or just have a built-in function to create a custom beep sound just like AHK... 

I assume that shouldn't actually be too difficult to implement in Python, mirroring the options for the AHK "beep" option, but then again as you know I'm no Python expert.

rbreaves:
@RedBearAK I could pretty easily add in some reliable sounds into the the Kinto installer at some point to make this sorta thing feasible. My installer already checks for many distros and can add in additional system level shortcuts that could execute a shell script.

But it would be better and more portable to let Python/xkeysnail handle it for sure...

RedBearAK:
@rbreaves 

Can't believe you didn't call me on the fact that this is a completely unworkable PR. 

It blocks the ability to use Backspace or Delete while renaming files. Duh. Because we still can't distinguish between the main app and text fields. Sure would be nice if we could. 

Somehow I managed to go through a whole day without noticing this, thinking I was being really clever. It turns out, I wasn't. 

Closing.

rbreaves:
Well I was already on the fence about it & some where in the back of my mind I was thinking "aren't there some instances where those keys really do get used?" and yea.. I wasn't about to go testing it right off to make sure at this very moment, but glad you found an error in your ways lol.

Have no idea how you would detect or know which part of the app you are in element wise without some x11 inspection going on and a rewrite of xkeysnail.. but I guess if someone what committed enough lol.

RedBearAK:
@rbreaves 

You said once or twice that you thought you had a way to detect when the keyboard focus is inside a text field, but you didn't elaborate. That would sure be handy for this and the whole "Enter key to rename, but then it won't work in text fields" problem. If I could solve that I'd be a happy camper, because I do a lot of file and folder renaming.

rbreaves:
That was when I had an IBUS related hack going on - I have no plans on re-implementing that at this time. Tbh it was a dirty hack when I did it pre xkeysnail any ways.

PR #610, opened 2022-01-21
Disable Duplicate (Cmd+D) shortcut in file managers

RedBearAK:
As much as I enjoy the convenience of the Cmd+D (Duplicate file/folder) shortcut, I have reluctantly made the decision that for most users it would be best to disable it by default, since it conflicts with the inherited function of "Add Bookmark" that Linux file managers like Nautilus have started using in parity with web browsers. Just one more thing that makes it easier to slide between web browsers and file "browsers". 

It's fun but has never been a "necessary" shortcut since a quick Cmd+C/Cmd+V will always accomplish the same thing. No one but me should miss it being active, and it will no longer interfere by default with bookmarking locations in Nautilus or any other file manager with that function assigned to Cmd+D (Ctrl+D). 

Anyone who really wants it can re-enable it.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -411,7 +411,7 @@ define_keymap(re.compile(filemanagerStr, re.IGNORECASE),{
     ###########################################################################################################
     K("RC-Super-o"):    K("RC-Shift-o"),        # Open in new window (or tab, depends on FM setup) [not universal]
     K("RC-Backspace"):  K("Delete"),	        # Move to Trash (delete)
-    K("RC-D"):          [K("RC-C"),K("RC-V")],  # Mimic Finder's Duplicate command (Copy, then Paste)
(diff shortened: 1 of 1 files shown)


rbreaves:
Well you're putting me in a strange position :p. But you are probably right. Trying to think of other examples where I went in another direction, none come to me right now, but I am sure there are some. Probably more on Windows than Linux.

I know recently I figured out how to properly unbind Win combos individually via regedit and that will certainly allow me to go back and set a few more things right as I was reluctant to just disable all default Win combos which is more widely documented on how to do.

RedBearAK:
@rbreaves 

Thanks. 

>I know recently I figured out how to properly unbind Win combos individually via regedit 

Oh boy, regedit. That sounds like "fun", of the opposite kind.

PR #624, opened 2022-02-16
Added missing libvte dependency

HelgeCPH:
The libvte dev library is missing when installing on Elementary OS (6.1).
Added its installation.

I do not know if this is the right place to introduce the installation of the
missing dependency. But I had to install it manually on Elementary OS 6.1
to make kinto start without a Python exception.

### DIFF
--- a/xkeysnail_service.sh
+++ b/xkeysnail_service.sh
@@ -305,6 +305,7 @@ fi
 
 if [[ $distro == "elementaryos" ]]; then
 	gsettings set io.elementary.terminal.settings natural-copy-paste false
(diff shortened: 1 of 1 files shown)


rbreaves:
Looks perfectly good to me! Although I am not certain if older versions use a different name. Doing version checking as well may be helpful too, but I will accept it regardless without checking into that. If this is what the latest version of elementaryOS requires then I will accept the change.

Figure the worst thing that can happen is it errors on older versions in the same way that it was for you on the latest. And this location looks good to me, where I would have located it.

Thanks!

Issue #622, opened 2022-02-11
Umlauts on Windows with american mac keyboard.


kirschem:
**Is your feature request related to a problem? Please describe.**
Using an american keyboard layout on macOS, I can type "umlauts" (ü,ä,ö) by typing OPT+U and then u, e or a.
This does not work with the current setup on Windows.

**Describe the solution you'd like**
Having the same behavior as described above.

**Describe alternatives you've considered**
It seems like the windows way of doing that would be AltGr + a/u/o, but that does not seem to work either (AltGr should be opt on mac, right?)

**Additional context**

kirschem:
Added a PR: https://github.com/rbreaves/kinto/pull/647

rbreaves:
Thanks for the PR! Keeping this open till I replicate this on the Linux side.

RedBearAK:
@rbreaves @kirschem 

Am I mistaken in thinking that this should be easily expandable, or at least replicatable, to work for pretty much all the other accented keys that are easily accessible in macOS with Option+Key1, Key2? Like the grave/acute accents, the enye, the "c" with the dangly bit, and so forth. Code seems pretty straightforward. 

On the other hand, I have this code already in my `kinto.ahk` since I just reinstalled Kinto on Windows 11, but it doesn't appear to be working as expected. It waits for the second key to be pressed, as it should, but then instead of a character with an umlaut I get a capital "A" with a tilde over it, and then a second character. Same goes for the other keys, just with different second characters. Like this: [ Ã¶ Ã– Ã¤ Ã„ Ã¼ Ãœ ]. Hope that comes out the way I'm seeing in my browser. 

I tried changing the "u" with umlaut to send "xyz" instead, and that works. It inserts "xyz" in Notepad when I type Opt+u, u. But the umlaut characters in the AHK config file seem to be incompatible with my (standard American, as far as I know) character set or something. I don't really know how that works in Windows. But to be universal it seems this might need to get AHK to output the characters in a different way. 

I'll keep playing with it for a bit.

RedBearAK:
[...]

PR #647, opened 2022-04-15
#622: Mirror umlaut behavior from macos with ahk

kirschem:
This aims to mirror the macos behavior for writing "Umlauts":

You can activate a modifier by hitting Option+u. If the next written character can be transformed into an "Umlaut", it's getting transformed.

Solves https://github.com/rbreaves/kinto/issues/622

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -858,3 +858,28 @@ Send {LWin up}
 Send {RShift up}
 Send {LShift up}
 return
+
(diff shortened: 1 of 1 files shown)


rbreaves:
Oh sweet & sorry I somehow missed respond to your original thread, but yea, I'll merge this. I will keep your original ticket open though until I replicate this behavior on the Linux half as well.

Issue #651, opened 2022-04-17
Finder Mods done for Windows File Explorer


RedBearAK:
@rbreaves 

I have the Finder Mods block for Windows File Explorer mostly done. I still want to get "Enter key to rename" working, but that will require more research and probably a lot of help from the AHK forums. All the important things are already working. I just need to know the best place to put the code in the Windows AHK config file. Right now I have it right after the `nxplayer.bin` block and before the "wordwise" stuff, but I don't know if that's the most logical place to put it. 

I was working on another block for the "Files" app alternate file manager, but because of how they are developing it the program doesn't have a proper EXE name, and the Win11 Explorer is so much simpler there's really not much point trying to support anything else. "Files" was the only usable alternative I could even find. So this is basically done. "Enter key to rename" can be added later if I can ever figure out how to make it work. 

Working shortcuts: 
- Get Info
- View Modes
- Nav with arrows*
- Nav with braces
- Open in New Window
- Safer delete shortcuts

*(special code was needed to stop folders opening in a new window with Cmd+Down)

Just let me know if you want this code in a different location in the `kinto.ahk` file and I'll put it in a PR at that location. Maybe right before the web browsers block, similar to the Linux config file? (Edit: No, it stopped working when I moved it there.)

Looks like this so far, and it's working nicely. You should be able to just paste this in your config and test it.  

```
#IfWinActive ahk_exe nxplayer.bin
    +F8::Send !+-     ; macOS TotalSpaces2 - Space Left
    +F6::Send !+=     ; macOS TotalSpaces2 - Space Right
#If
; +F8::Send {LCtrl down}{LWin down}{left}{LCtrl up}{LWin up}  ; Comment out on host machine
; +F6::Send {LCtrl down}{LWin down}{right}{LCtrl up}{LWin up} ; Comment out on host machine

; #########################################################################
; #############   START OF FINDER MODS FOR FILE MANAGERS   ################
; #########################################################################
; Finder Mods for Windows File Explorer (explore.exe)
#IfWinActive ahk_class CabinetWClass ahk_exe explorer.exe
    ^i::Send !{Enter}           ; Cmd+i: Get Info / Properties
    ; Enter::Send {F2}          ; Enter: Rename file/folder (FIX: NEEDS TO WORK INSIDE TEXT FIELD WHILE RENAMING)
    ^r::Send {F5}               ; Cmd+R: Refresh view (Not actually a Finder shortcut? But works in Linux file browsers too.)
[...]

PR #652, opened 2022-04-18
[Win] Finder Mods for Windows File Explorer file manager

RedBearAK:
As mentioned in issue #651, this works for me, but if it isn't in the best location in the file just move it or do your own commit and then close this, or let me know if you've merged the code separately so I can close this. 

Working shortcuts in this PR to get Windows File Explorer (explorer.exe) to behave more like macOS Finder: 
- Get Info (Cmd+i)
- View mode shortcuts (Cmd+1/2/3/4)
- Nav with arrows, including code that stops Explorer opening new windows for Cmd+Down on folders
- Nav with braces (prior/next in history)
- Open in New Window (Cmd+Shift+o)
- Safer delete shortcuts, with error beeps for Backspace or Delete used without Cmd key
- Enter key to rename like Finder (without blocking other usage of Enter key)

Code passes Enter, Backspace and Delete keys when focus is in editable text input fields (and passes Enter key in sidebar for navigation).

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -214,6 +214,58 @@ GroupAdd, intellij, ahk_exe idea64.exe
 ; +F8::Send {LCtrl down}{LWin down}{left}{LCtrl up}{LWin up}  ; Comment out on host machine
 ; +F6::Send {LCtrl down}{LWin down}{right}{LCtrl up}{LWin up} ; Comment out on host machine
 
+; #########################################################################
+; #############   START OF FINDER MODS FOR FILE MANAGERS   ################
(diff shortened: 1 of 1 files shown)


rbreaves:
Tested. Thanks, sorry it took so long.

RedBearAK:
@rbreaves 

Thanks! 

I have been hearing rumors that they had actually had tabs in Explorer for a while in Win11, but they were removed, and fairly soon they will be coming back in a new Win11 revision. So I'll need to add another PR related to this when the tabs come back and I can test the relevant shortcuts.

PR #666, opened 2022-05-19
Add Deepin Browser to browsers and chromes WM_CLASS lists

RedBearAK:
Adding the WM_CLASS for Deepin Browser (org.deepin.browser) to the general browsers and the Chromes WM_CLASS lists. Browser is based on Chromium.


rbreaves:
It's the devil, but I will allow it.

RedBearAK:
LOL.

PR #686, opened 2022-06-07
(fix) small sleep after  menu opens between hitting S

joshgoebel:
Without this delay sometimes the input will go into the active webpage BEFORE the menu opens...

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -444,7 +444,7 @@ define_keymap(re.compile("Firefox", re.IGNORECASE),{
 })
 
 define_keymap(re.compile(chromeStr, re.IGNORECASE),{
-    K("C-comma"): [K("M-e"), K("s"),K("Enter")],    # Open preferences
+    K("C-comma"): [K("M-e"), sleep(10/1000), K("s"),K("Enter")],    # Open preferences
     K("RC-q"):              K("M-F4"),              # Quit Chrome(s) browsers with Cmd+Q
     # K("RC-Left"):           K("M-Left"),            # Page nav: Back to prior page in history (conflict with wordwise)
     # K("RC-Right"):          K("M-Right"),           # Page nav: Forward to next page in history (conflict with wordwise)


RedBearAK:
> Without this delay sometimes the input will go into the active webpage BEFORE the menu opens...

@joshgoebel 

Are you sure this works with the Kinto branch? I tried inserting a `sleep` in a macro and it just causes the macro to stop at that point. 

When I tried to use `sleep` in the past I discovered that the `sleep` was being evaluated and taking effect when the config file was being loaded, not when the shortcut is triggered. 

See this thread of mine on `mooz/xkeysnail`. 

https://github.com/mooz/xkeysnail/issues/149

Those `launch_and_then` and `sleep_and_then` functions came from that thread. Using those modified functions was the only way I could actually get a `sleep` delay to work as part of a shortcut.

joshgoebel:
> Are you sure this works with the Kinto branch? 

No, but I'm happy if we hold it until we have a change to more thoroughly test with keyszer.  I honestly don't see how it worked before just glancing at the 0.4.0 source... I had to make some changes to how commands are handled to get it working with my fork... I guess I just assumed if it was in the README it should work...

RedBearAK:
>I honestly don't see how it worked before just glancing at the 0.4.0 source... I had to make some changes to how commands are handled to get it working with my fork...

I have a feeling it never really did, at least not as intended. If you fixed it in your branch, that's great. But unless those changes make it into the Kinto branch of `xkeysnail`, adding the `sleep` to the Kinto config file right now will probably just break things.

joshgoebel:
Actually looking right now at kinto's xkeysnail it looks like it should work.

RedBearAK:
@joshgoebel 

Having trouble with the latest iteration of `keyszer`. Even tried deleting and re-cloning from scratch. 

```
keyszer v0.4.99
Traceback (most recent call last):
  File "/home/kris/Downloads/josh-xkeysnail/xkeysnail/./bin/keyszer", line 6, in <module>
    main()
  File "/home/kris/Downloads/josh-xkeysnail/xkeysnail/src/keyszer/cli.py", line 96, in main
    eval_config(args.config)
  File "/home/kris/Downloads/josh-xkeysnail/xkeysnail/src/keyszer/cli.py", line 16, in eval_config
    exec(compile(config_code, path, 'exec'), globals())
  File "/home/kris/.config/kinto/kinto.py", line 10, in <module>
    
ModuleNotFoundError: No module named 'xkeysnail'
```

Going to be changing the repo name at some point?

joshgoebel:
The name changed. Need to update your configure file imports.

joshgoebel:
The last line of the trace shows you where the error is.

RedBearAK:
@joshgoebel 

>The last line of the trace shows you where the error is.

Ah. That means I need to either use a separate config file now for testing with `keyszer`, or figure out how to use an optional import setup that "tries" to do both imports and fails more gracefully for the one that's missing. 

This makes the process of switching back and forth more complicated. I'll also have the update my "stop" scripts to make sure it kills any instance of `keyszer` along with `xkeysnail` processes, to avoid conflicts. Usually I kill it in the terminal before starting the other one, but sometimes I forget. 

Well, I'll get to it, I guess.

joshgoebel:
With Keyszer you don't need any imports at all

RedBearAK:
@joshgoebel 

>With Keyszer you don't need any imports at all

Oh. So if I can just remember how to make the `xkeysnail` import optional, I should be good to use the same config file between the two?

joshgoebel:
I think so.

RedBearAK:
@joshgoebel 

Couldn't manage to replicate what I did in my `play_sound.py` script. Just commenting out the line makes it work with `keyszer`. 

But, now the tab in VSCode says "1000 problems with this file". Due to the lack of an import defining the functions, I'm assuming. 

I'm used to VSCode only showing warnings if I've actually missed some syntax or something, so I consider this state detrimental to the ability to troubleshoot the config file. Does this not affect the IDE you happen to be working in? 

Just noticed that I also can't Shift-select text with the mouse, without holding Shift past the suspension time. I suppose I should have expected that.

joshgoebel:
> Just noticed that I also can't Shift-select text with the mouse, without holding Shift past the suspension time. I suppose I should have expected that.

Did you not take the steps I showed you to get your touchpad into flowing thru?  Once you do this should all "just work".

joshgoebel:
> I'm used to VSCode only showing warnings if I've actually missed some syntax or something, so I consider this state detrimental to the ability to troubleshoot the config file. Does this not affect the IDE you happen to be working in?

I don't need the linter, if there is an error I'll see it when I run the config... but not needing them is just a conveneience... you can still import if you really want to - just it's now not necessary for a cleaner file.

RedBearAK:
@joshgoebel 

>I don't need the linter, if there is an error I'll see it when I run the config... but not needing them is just a conveneience... you can still import if you really want to - just it's now not necessary for a cleaner file.

It's just a convenience for your environment, but in my environment it means my entire file is now lit up with squiggly lines and yellow errors. The sidebar preview area looks like a Christmas tree. 

I think for the `keyszer` import to stop this from happening, VSCode has to be able to "find" the module, which of course it can't unless it's installed outside the `venv`. I guess that's safer now that the name is different... 

>Did you not take the steps I showed you to get your touchpad into flowing thru? Once you do this should all "just work". I've been modifier clicking the past two days once I routed the mouse data thru.

I kind of thought you were just throwing out ideas and would integrate those changes into your code in a generic fashion, if they are necessary to make it fully work. 

My intention here was always to be like more of an end user willing to test pre-release software and give feedback. Which I've done to the best of my ability. You've seen that my skills with python are pretty weak. I'm not clear on where that code was even supposed to go, or whether it needed to be customized somehow to my devices. Or, in turn, whether it would get overwritten the next time I did a `git pull`.

joshgoebel:
> but in my environment it means my entire file is now lit up with squiggly lines and yellow errors.

As I said, it's your choice - if you want to include them because they are beneficial, include them.  That's just fine. :)

> VSCode has to be able to "find" the module, which of course it can't unless it's installed outside the venv

Sorry I don't know about this, my kinto.py is local in my project... I'd just copy yours over for testing (or make a symlink).

RedBearAK:
@joshgoebel 

I'll just brute force it for now. I don't want to deal with two different config files. 

```
cat ~/.config/kinto/kinto.py | grep -v "from xkeysnail" > kinto_config.py

./bin/keyszer --watch -c kinto_config.py
```

But now I'm seeing strange problems with stuck keys. Like getting stuck with the task switcher dialog held open if I use Cmd+Shift+Tab. I'll try to describe it in a `keyszer` issues thread.

RedBearAK:
@joshgoebel 

Completely forgot why I was trying to run `keyszer` again, to test the `sleep` in a macro: 

Fails in Kinto branch, works in `keyszer`. It makes the macro stop at the point where you insert the `sleep` with the Kinto branch. 

```
Kinto: 
Now is the time for all good men

keyszer:
Now is the time for all good men to come to the aid of their country.
```

The macro: 

```
    K("RC-Shift-M-t"):          [K("Shift-N"),K("o"),K("w"),K("Space"),K("i"),K("s"),K("Space"),K("t"),K("h"),K("e"),K("Space"),
    K("t"),K("i"),K("m"),K("e"),K("Space"),K("f"),K("o"),K("r"),K("Space"),K("a"),K("l"),K("l"),K("Space"),K("g"),K("o"),K("o"),
    K("d"),K("Space"),K("m"),K("e"),K("n"),sleep(1),K("Space"),K("t"),K("o"),K("Space"),K("c"),K("o"),K("m"),K("e"),K("Space"),K("t"),K("o"),
    K("Space"),K("t"),K("h"),K("e"),K("Space"),K("a"),K("i"),K("d"),K("Space"),K("o"),K("f"),K("Space"),K("t"),K("h"),K("e"),K("i"),
    K("r"),K("Space"),K("c"),K("o"),K("u"),K("n"),K("t"),K("r"),K("y"),K("dot"),K("Enter")
```

joshgoebel:
Well, weird... so maybe we should leave this open as a bug report though? Thoughts?

RedBearAK:
@joshgoebel 

If you know how to fix it in the Kinto branch of `xkeysnail` I'd make a PR over on `rbreaves/xkeysnail` and see if Ben will accept it, and I'd probably make a more specific issue post for Kinto that points to the `xkeysnail` PR. 

If you close this PR for now you can always re-open it later when Kinto is patched to make it work, and you know it won't break things. That's what I would do, I think. 

I haven't checked so I don't know if this bug affects mainline `xkeysnail` or just the Kinto branch.

joshgoebel:
> If you know how to fix it in the Kinto branch of xkeysnail 

My time is likely better spent on my fork.  I'll close this for now.

PR #687, opened 2022-06-07
Support VS Code without MS pieces

joshgoebel:
Adds `code-oss` which is the open source VS Code without the Microsoft phone home pieces.


joshgoebel:
Is there anything I can do to help get some of these tiny (and surely unobjectionable) PRs moving?

PR #692, opened 2022-06-09
Replace M, LM, RM with Alt, LAlt, RAlt

joshgoebel:
Removes any ambiguity because Key.*META is NOT Alt, it's (according to our own mapping!) Super.

Long-term `M` should probably go away (to further remove ambiguity or be replaced with META.

Context: https://github.com/mooz/xkeysnail/issues/158

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -185,16 +185,16 @@ define_conditional_modmap(re.compile(termStr, re.IGNORECASE), {
 # Keybindings for IntelliJ
 define_keymap(re.compile("^jetbrains-(?!.*toolbox).*$", re.IGNORECASE),{
     # General
-    K("C-Key_0"): K("M-Key_0"),                 # Open corresponding tool window
-    K("C-Key_1"): K("M-Key_1"),                 # Open corresponding tool window
-    K("C-Key_2"): K("M-Key_2"),                 # Open corresponding tool window
-    K("C-Key_3"): K("M-Key_3"),                 # Open corresponding tool window
-    K("C-Key_4"): K("M-Key_4"),                 # Open corresponding tool window
-    K("C-Key_5"): K("M-Key_5"),                 # Open corresponding tool window
-    K("C-Key_6"): K("M-Key_6"),                 # Open corresponding tool window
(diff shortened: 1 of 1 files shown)


joshgoebel:
Can this please get merged before I start work on the Mac left side retooling?

rbreaves:
Merged! Thanks @joshgoebel Sorry it took me a little while to get it merged.

linux/kinto.py
```diff
@@ -190,16 +190,16 @@
...
+    K("C-Key_2"): K("Alt-Key_2"),                 # Open corresponding tool window
+    K("C-Key_3"): K("Alt-Key_3"),                 # Open corresponding tool window
+    K("C-Key_4"): K("Alt-Key_4"),                 # Open corresponding tool window
+    K("C-Key_5"): K("Alt-Key_5"),                 # Open corresponding tool window
+    K("C-Key_6"): K("Alt-Key_6"),                 # Open corresponding tool window
+    K("C-Key_7"): K("Alt-Key_7"),                 # Open corresponding tool window
+    K("C-Key_8"): K("Alt-Key_8"),                 # Open corresponding tool window
+    K("C-Key_9"): K("Alt-Key_9"),                 # Open corresponding tool window
     K("Super-Grave"): K("C-Grave"),             # Quick switch current scheme
     K("C-Comma"): K("C-M-s"),                   # Open Settings dialog
```
naveg, line 204: Do places like this where `M` occurs in the middle of a chord need to be updated as well?

RedBearAK: @naveg @rbreaves @joshgoebel 

Yes, I think someone needs to do a find and replace for all the `-M-` or `-LM-/-RM-` instances in the Kinto config. I believe I did that manually on my own config file back when I converted `M` to `Alt`. I can't find any instances of this in mine. 

I think this won't actually have any negative effect until Kinto switches over from `xkeysnail` to `keyszer`, since `xkeysnail` is still treating `M` as `Alt`. But it definitely needs to be fixed before the switchover.

naveg: https://github.com/rbreaves/kinto/pull/776

PR #698, opened 2022-06-11
A proper ignore_key helper

joshgoebel:
I suppose you'd want me to fix the spacing? :)


joshgoebel:
Doesn't fix the bug, just makes your file more readable.

joshgoebel:
_Worth noting you already use None in several places to do exactly this, ignore keystrokes._

PR #719, opened 2022-06-18
[Win] Option key special character entry scheme

RedBearAK:
This is an AHK implementation of the full selection of special characters, symbols, and dead keys accented characters that are available on the standard US Apple keyboard using Option or Option+Shift. Uses the AHK Unicode method, which works even with applications like Notepad that fail with the Alt Code character entry method. 

The Option key scheme will interfere with Alt and Alt+Shift shortcuts while active, so it is disabled by default when the AHK script is loaded. It can be enabled and disabled again with Shift+Option+Command+O (the letter "o"), or by using the new tray menu item. The tray menu item will receive a check mark when the scheme is active. 

Alert messages are presented to the user on enable and disable. The enable MsgBox warns about the conflict and reminds the user how to disable the scheme. 

Known issues: 

Option+g will work, but will bring up Windows XBox "Game Bar". This can be disabled in PowerShell: 

```
Get-AppxPackage Microsoft.XboxGamingOverlay | Remove-AppxPackage
```

Typing Option+L will work, but will also lock the screen since the keys are Win+L. This can only be disabled in the registry, which will completely disable the ability to lock the screen. There's no known better solution. But it's very quick to log back in, so it's no big deal. 

The Apple logo (Shift+Option+K) exists only in the Baskerville Old Face font, and the related shortcut will display a MsgBox explaining how to make it display properly. 

All other characters work without known issues. 

Documentation of all the included characters (137 different characters, if my math is right) that can be typed when this scheme is enabled is here: 

https://github.com/RedBearAK/optspecialchars/blob/main/Option-key-characters.md

This PR replaces and enhances the Umlaut function submitted by user Kirschem in PR #647.

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -906,46 +906,20 @@ toggle_optspecialchars:
         StringCaseSense, On
         ; watch next input string
         Input, UserInput, L1 
-        ; Watch for Escape key, cancel dead keys sequence
-        if UserInput = Esc
-            Return
-        ; Option/Alt key needed to access menus sometimes
-        ; If user repeats same shortcut, just send to app unmodified.
-        ; else if UserInput = !SC029
-        ;     Send, {Alt down}{SC029}{Alt up}
-        else if UserInput = a
-            ; à {U+00E0} (Alt+0224)
-            Send, {U+00E0}
-        else if UserInput = e
-            ; è {U+00E8} (Alt+0232)
-            Send, {U+00E8}
-        else if UserInput = i
-            ; ì {U+00EC} (Alt+0236)
-            Send, {U+00EC}
-        else if UserInput = o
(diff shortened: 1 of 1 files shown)


RedBearAK:
@rbreaves 

On the advice of someone in the AHK forums, I converted the if-else logic blocks for the "dead keys" to Switch/Case commands.

rbreaves:
Really amazing work! I appreciate the PR, just got done testing it. Merging it now!

RedBearAK:
@rbreaves 

> Really amazing work! I appreciate the PR, just got done testing it. Merging it now!

Thanks a lot! Will be trying to do the same thing on the Linux side, with the assistance of `keyszer`, since it needs nested keys and possibly a new function to enable/disable with a shortcut.

PR #722, opened 2022-06-22
[Win] Optional fix for media arrow keys

RedBearAK:
Apple laptop keyboards have the PgUp/PgDn/Home/End navigation functions on Fn+arrow keys (even if not explicitly labeled on the keys). Most PC laptops also have these functions on the arrow keys, but some PC laptop keyboards have media functions (Play_Pause/Stop/Prev/Next) on the arrow keys instead. This makes it impossible to use the Fn+arrow keys for shortcuts like text selection (with Shift), or general navigation. 

This optional fix for this issue (disabled by default) will cause media arrow keys to behave like an Apple keyboard when used with the Fn key. All possible modifier combinations are mapped to what they would have been without Kinto's modifier remap. 

This PR creates a new tray menu item that will toggle the fix on and off, and it can be toggled with Shift+Opt+Cmd+M. The tray menu item will update with a check mark when the fix is active, and a MsgBox alerts the user when it is toggled on and off. 

If desired the user can set the fix to be enabled by default by changing a single variable value near the top of `kinto.ahk`. The toggles to disable/enable will still function. 

Companion to part of PR #499 which is for Linux.

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -62,6 +62,8 @@ Menu, Tray, Add, Autodetect Keyboards, autodetect
 ; Menu, Tray, check, Autodetect Keyboards ; Autodetect
 ; Menu, Tray, disable, Autodetect Keyboards ; CB/IBM
 Menu, Tray, Add, Suspend Kinto, tray_suspend
+; Add tray menu item for toggling media_arrows_fix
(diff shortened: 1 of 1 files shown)


RedBearAK:
Closing in favor of PR #724, which does the same thing but smarter, with fewer lines, using a `modmap` style permanent redirect of the base media key.

PR #725, opened 2022-06-24
Staging modmap for toggle-able Media Arrows Fix

RedBearAK:
This is a more elegant solution to the need to fix some PC laptop or compact keyboards with media functions on Fn+arrow keys to have PgUp/PgDn/Home/End nav keys on the Fn+arrows, like an Apple laptop keyboard. 

Disabled modmap lines, meant to be optionally enabled by the user, manually or with a script or tray menu tweak item (future). All lines tagged with `(media_arrows_fix)`.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -148,6 +148,13 @@ define_conditional_modmap(lambda wm_class: wm_class.casefold() not in terminals,
     # Key.LEFT_CTRL: Key.LEFT_META,   # Mac
     # Key.RIGHT_META: Key.RIGHT_CTRL, # Mac - Multi-language (Remove)
     # Key.RIGHT_CTRL: Key.RIGHT_META, # Mac - Multi-language (Remove)
+    
+    # Fix arrow keys with media functions instead of PgUp/PgDn/Home/End
+    # Key.PLAYPAUSE:      Key.PAGE_UP,        # (media_arrows_fix)
+    # Key.STOPCD:         Key.PAGE_DOWN,      # (media_arrows_fix)
+    # Key.PREVIOUSSONG:   Key.HOME,           # (media_arrows_fix)
+    # Key.NEXTSONG:       Key.END,            # (media_arrows_fix)
+
 })
 
 # [Conditional modmap] Change modifier keys in certain applications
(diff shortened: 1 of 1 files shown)


rbreaves:
Which laptops have media key functions on the arrow keys!?

RedBearAK:
> Which laptops have media key functions on the arrow keys!?

Don't know about others yet, but my two Acers do. 

They probably didn't have room to put the media functions above the Numpad because it's an odd 3-column Numpad with no room for extra keys. Function key row is populated with a bunch of hardware related controls. Display, touchpad on/off, Wi-Fi, sleep. Since the Numpad has nav keys I guess they figured why not put media functions on the arrows?

rbreaves:
I wonder if we should specify in the comments that this may be specific to Acer laptops, so there's some understanding that this isn't necessarily applicable to other laptops.

RedBearAK:
> I wonder if we should specify in the comments that this may be specific to Acer laptops, so there's some understanding that this isn't necessarily applicable to other laptops.

I’m sure there are other PC laptops that have made the same choice. I just don’t know a specific one to point to yet. It should be fairly obvious to the user whether they need this fix or not, when they look at their arrow keys. 

This would be applicable to any keyboard that chose to put media functions on the arrows. There may even be a compact external keyboard somewhere that chose to do the same thing. I don’t see any benefit to going out of our way to say it may only apply to certain Acer laptops.

rbreaves:
My old thinkpad would disagree. I’d just rather have an idea of what brand(s) it’s most likely associated w/. I get specific models are a bit much to ask though.

BF28BA09-B5E3-4F73-9753-90ECFA0AE19F

Typically none of my laptop have media keys on those keys at all. Also if something is designed later to try & enable or disable hotkeys based on brand, make or model then that info would be helpful to have.

RedBearAK:
> My old thinkpad would disagree.

Oh, nice. So you do have one with media functions on arrows. 

This fix feels to me like not something that should necessarily be automatically enabled during install. It would be too hard to track down the models that might be affected. But you could just ask the user during install if they have a keyboard with media functions instead of PgUp/PgDn/Home/End on the arrow keys, and if so, ask if they want to enable the fix for the issue. 

I did try to add an item to the Tweak dialog in the past, but it was far more complicated than it was on the AHK side. But that's where it belongs. 

My Acers are A515-43 variants, if you really want to start adding models to a list that you'll just automatically fix one day. Nice idea. 

> I still plan to implement F row media keys at some point for MacBooks based on their specific models so that’s part of why I’m thinking along these lines.

I'm a bit confused by this. On my MacBooks that I've tried with Linux, those keys that do media functions and volume and brightness all work fine. At least I thought they did, unless only the volume and brightness controls were working as expected.

RedBearAK:
Media keys are definitely working OK on 2007 white MacBook running Fedora 36 KDE Spin.

rbreaves:
I haven’t been using the media keys much under Linux so I wasn’t entirely sure 😅. So yea may be a non-issue or just users w/ very good memory that want the same media key layout regardless of their laptop. In which case they’ll want a drop down & preview.

PR #726, opened 2022-07-04
[Win] Typo/format cleanup in optspecialchars

RedBearAK:
No substantive changes. Just fixing some inconsistent formatting and missing special character examples in comments for the Option-key special character entry scheme. All comments in that section should now be normalized to the same pattern, with a reference sample for each supported character. 

```
; Description {UNICODE}: SAMPLECHAR (Alt Code)
```

If `(Alt Code)` is missing it's because no Alt Code was found for that character.

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -943,7 +943,6 @@ toggle_optspecialchars:
     }
     return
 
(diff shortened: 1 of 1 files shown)


RedBearAK:
Closing in favor of PR #731.

PR #731, opened 2022-07-15
[Win] Enhance OptSpecialChars dead keys processing

RedBearAK:
Adds more accurate and intuitive behavior to the process of using dead keys, to match macOS more closely, and fixes some of the Switch/Case lines that weren't working as expected (Escape, Space were not actually activating). 

Dead keys sequences will now place the pending accent diacritic character in a highlight, then either replace it with an accented character if there is a match, or deselect the diacritic character and exit the dead keys sequence, leaving the diacritic character visible. Any input other than Escape or Space or a matching letter will be passed through. This is a match for how dead keys work in macOS. 

Includes the updated comments and typo fixes, so I'll close PR #726.

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -943,7 +943,6 @@ toggle_optspecialchars:
     }
     return
 
-; #IfWinNotActive ahk_group remotes
 #If !WinActive("ahk_group remotes") && optspecialchars = 1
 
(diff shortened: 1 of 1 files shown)


RedBearAK:
Closed due to some previously unnoticed behavior from editing keys that don't generate characters, and capital accented characters not working correctly. Reworked and replaced with PR #732.

PR #732, opened 2022-07-15
[Win] Enhance OptSpecialChars dead keys processing

RedBearAK:
For Windows/AHK: 

Adds more accurate and intuitive behavior to the process of using dead keys, to match macOS more closely, and fixes some of the Switch/Case lines that weren't working as expected (Escape, Space were not actually activating).

Dead keys sequences will now place the pending diacritic accent character (e.g., umlaut [ ¨ ]) in a highlight, then either replace it with the full accented character (e.g., [ Ü ]) if there is a match, or deselect the diacritic character and exit the dead keys sequence, leaving the diacritic character visible. Any input other than Escape or Space or a matching letter will be passed through. This is a very close match for how dead keys work in macOS.

Also enabled the ability to interrupt a dead key sequence by triggering another dead key sequence, which like many other keys will just leave the original diacritic character visible and insert the new one in a highlight as the start of the new sequence. This behaves like it does on a Mac when you change your mind in the middle and do a different dead key. 

Previous PR by this name had some unforeseen issues with non-character-generating input editing keys like arrows, Backspace, and Delete, which were being ignored/disabled by the AHK `Input` command. And the uppercase accented characters were not working properly. All such issues are fixed here and thoroughly tested. 

🤞🏽 😮‍💨 🤞🏽

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -943,7 +943,6 @@ toggle_optspecialchars:
     }
     return
 
-; #IfWinNotActive ahk_group remotes
 #If !WinActive("ahk_group remotes") && optspecialchars = 1
 
     ; ######   NUMBER KEYS ROW   ######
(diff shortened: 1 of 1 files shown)


RedBearAK:
Added support for exiting dead keys with appropriate actions for the arrow keys, mimicking what happens in macOS as closely as possible.

Danilo-Araujo-Silva:
@RedBearAK, do you know if these modifications would also work for this configuration?

I'm having a hard time with a Magic Keyboard here on KDE ^^
Without the dead keys working is really annoying..

```
Operating System: Kubuntu 22.04
KDE Plasma Version: 5.24.7
KDE Frameworks Version: 5.92.0
Qt Version: 5.15.3
Kernel Version: 6.2.0-37-generic (64-bit)
Graphics Platform: X11
Processors: 8 × Intel® Core™ i7-6820HQ CPU @ 2.70GHz
Memory: 15.4 GiB of RAM
Graphics Processor: Mesa Intel® HD Graphics 530
```

Danilo-Araujo-Silva:
@RedBearAK, thanks a lot for your quick and informative reply :)

I'm going to check your repo here o/ Uhuul

PR #744, opened 2022-08-20
Mac numpad feature

RedBearAK:
Additions to both the Windows and Linux config files to enable the feature of making the numpad behave like a real Mac keyboard. Numbers are always numbers, Numlock key becomes "Clear", which sends Escape. 

Feature can be disabled and re-enabled with Option+Numlock. Windows version adds a menu item with a check mark. 

Notifications or dialogs will appear on either platform to indicate that the feature has been enabled or disabled when using the shortcut. 

Since the goal of Kinto is to mimic an Apple keyboard, I will argue that it makes sense to have this feature enabled by default. The user can disable it by changing a variable value in the config file.


joshgoebel:
Sounds awesome.

PR #776, opened 2022-12-12
Replace more occurences of M with Alt

naveg:
A few of these were missed in #692, specifically occurences in the middle of chords.

Follow up to https://github.com/rbreaves/kinto/pull/692#discussion_r1045144539


RedBearAK:
@naveg 

Some of the output combos still have `-M-`.

naveg:
Good catch, fixed. Search and replace is hard apparently :upside_down_face:

Issue #787, opened 2023-02-04
Delete button in Windows Explorer not working


andyd0:
I updated Kinto recently and forgot that it overrides my config file. Anyone able to help with getting the delete button working again?

RedBearAK:
@andyd0 

✋🏽 Hi, this is my fault. Although I did this on purpose, and I'll explain. These are due to the "Finder Mods". 

Disclaimer: I am not the dev, just a contributor of this particular part of the config. 

Since Kinto's purpose in general is to mimic the keyboard behavior in macOS applications, for people who like how macOS works, I implemented some shortcut overrides for Windows Explorer to mimic Finder behavior. I've done the same thing for several different common Linux file managers (or as they are often called in Linux, file browsers). 

The Enter key in the Finder has always been used for renaming, which a lot of Mac users like me got used to using over the much more awkward F2 shortcut that Linux and Windows uses. The F2 shortcut is often a real pain on laptops especially, where you need to go find and press the Fn key before F2, since there are typically special hardware functions on the F-keys row on laptops. Mine has a hardware function to turn Wi-Fi off on the F2 key. So in Windows users tend to use the much slower right-click menu to rename things in Explorer. I find this a huge pain ever since I got used to how it works in the Finder. Enter-rename-Enter, done. It's super fast. 

To actually open files or run applications (via the keyboard) in the Finder, you use the same shortcut as you use to "drill down" into (or "open") folders in the Finder: Cmd+Down (arrow). The opposite of going "up" to the parent folder with Cmd+Up. I find it quite logical to think of it this way. 

Delete (and Backspace) are blocked from deleting files immediately, in favor of the Cmd+Backspace (aka, the bigger Delete key on an Apple keyboard) shortcut used in the Finder. Because that's how all these keys work in the Finder. 

### How to disable the Finder Mods

If you never got used to using these keyboard shortcuts in the Finder on macOS, and you really dislike them, you can disable them pretty easily. Just open your `kinto.ahk` config file and search for "finder", you'll see this line just below it: 

```
#IfWinActive ahk_class CabinetWClass ahk_exe explorer.exe
```

If you change the application name to something that won't match, the whole set of "Finder Mods" should stop working as soon as you close and restart Kinto (you can restart Kinto without logging out by double-clicking on the kinto-start.vbs file in the config folder). 

Example: 

```
#IfWinActive ahk_class CabinetWClass ahk_exe NOTexplorer.exe
```

[...]

PR #795, opened 2023-04-02
Make Finder Mods active on desktop, block wordwise shortcut

RedBearAK:
This will make the Finder Mods shortcuts active on the Windows desktop (which has a different `ahk_class` than the usual Explorer windows). Also, this will block the improper action of the "wordwise" shortcut Cmd+Backspace while on the Windows desktop, unless the user is editing a file/folder name on the desktop. The wordwise shortcut was doing "select all icons to top-left corner, delete". Cmd+Backspace should still work when in a "text edit" kind of area, like when renaming files/folders on the desktop.

Should resolve an issue brought up in #787 comment thread.


RedBearAK:
Closing because this is superseded by the fixes in PR #867.

PR #828, opened 2023-09-04
Added Keybindings for Gedit

satyapatel2004:


RedBearAK:
@satyapatel2004 

```py
#Keybindings for for GEDIT, the Gnome Text Editor: 
define_keymap(re.compile("^gedit$", re.IGNORECASE), {
    K("Super-c") : K("C-c"), #Copy 
    K("Super-v") : K("C-v"),  #Paste
    K("Super-s") : K("C-s") #Save 
}) 
```

I'm not sure this is something the Kinto dev will want to include in the default config file. You appear to be remapping the physical Ctrl key (logically mapped to Super outside of terminals) back to Ctrl logically, so you can Cut/Copy/Save the same way you normally would in Linux or Windows. But why? With Kinto's remapping those functions are done with the modifier key next to the Space bar simulating the Cmd key. The point of Kinto is to make everything work like macOS. 

The closest equivalent app in macOS is TextEdit, and those physical key shortcuts don't perform those actions in TextEdit. 

This is a type of personal customization that probably doesn't make sense to add to the default Kinto config file. Nobody coming from macOS would be expecting those physical shortcuts to perform those actions in Gedit.

rbreaves:
I mean yea.. isn't this just remapping the base remaps and sorta breaking the universality of using Cmd convention for basically everything. Does not make much sense to me remapping Cmd/Ctrl back to the Super key which would be in the traditional Ctrl position during GUI app usage.

PR #838, opened 2023-11-20
add chrome and finder shortcut

yasapurnama:


RedBearAK:
@yasapurnama 

I'm not the Kinto dev, just a minor contributor. I am the author of the "Finder Mods" keymaps for Linux file managers (and the stuff in the Windows version of Kinto focused on Windows Explorer). 

### About the Linux shortcuts

The hidden files shortcut should already work in Thunar, I just tested it on Fedora 39. 

So this shortcut: 

```py
    K("C-Shift-DOT"):   K("LC-H"),                  # Show/Hide Hidden Files
```

Indicates to me that you may be using the physical Ctrl key instead of the key next to your Space bar, which should work fine for this. Remember the goal of the config is to use the physical keys that would perform the same action in macOS. 

There is an existing shortcut that applies to all the file managers in the list, including Thunar. Notice that it uses "RC" instead of just "C" on the input side, so that it only applies to the key next to the Space bar. 

```py
    K("Shift-RC-dot"):          K("RC-H"),                      # Show/hide hidden files ("dot" files)
```

This one is probably good, but it should use "RC" instead of "C": 

```py
K("C-M-Backspace"): K("Shift-Delete"), # Delete Files Permanently
```

And I should probably look at setting a general file manager shortcut for that. I think I didn't bother because that's a potentially very dangerous shortcut (deleting things without sending to the trash). 

In this set of shortcuts, there are a couple of things to note: 

```py
    K("RC-Y"): K("RC-H"),           # Browser History
    K("M-RC-U"): K("RC-U"),         # View Page Source
    K("M-RC-L"): K("RC-J"),         # Downloads
```

1. There's generally no reason to use any right/left specific modifier references on the right side of the colon (the output side), since apps won't usually react differently to right vs left modifiers (there are very few exceptions). But on the left side, as noted, the "RC" limits in the input combo to just the key next to the Space bar, which becomes "RIGHT_CTRL", even on an Apple keyboard. 

2. The "M" reference to the Alt key, as if it was the "Meta" key, was always wrong compared to what the Linux kernel refers to as the "Meta" key (the Super/Win/Cmd key), and has been deprecated in the keymapper fork that Kinto will probably be moving to at some point. Any usage of "M/RM/LM" as a modifier should be replaced with "Alt/RAlt/LAlt" so it would keep working in the future. That's already been done throughout the rest of the default Kinto config. 

These seem good otherwise, if they are really the correct input shortcuts for Chrome(s) on macOS. And... No, one of them seems to be incorrect. The latest Chrome on a Big Sur machine I have uses Shift+Cmd+J for opening the browser's Downloads view. So that one should be: 

```py
    K("Shift-RC-J"): K("RC-J"),         # Downloads
```

But the others are correct, and missing from the default config. 

### About the Windows shortcuts

This is an interesting one, and I wonder if it works the same way on Windows 10 and 11. Probably, but the Windows Explorer has been changing a lot lately. 

```ahk
^+.::send {Alt Down}vhh{Alt Up} ; Show/Hide Hidden Files
```

This one, again, is a potentially very dangerous shortcut that could end up deleting entire folder structures without sending them to the trash, and I'm not sure it is wise to have it in the default config. But that will be up to the Kinto dev. 

```ahk
$^!Backspace::send {LShift down}{Delete}{LShift up} ; Delete Files Permanently
```

Same note about this shortcut to open the Chrome Downloads view: 

```ahk
!^l::send {Ctrl Down}j{Ctrl Up}
```

Should be: 

```ahk
+^j::send {Ctrl Down}j{Ctrl Up}
```

Since the goal of Kinto is to use the current macOS shortcut for that function in the matching app on macOS. 

I don't know if that shortcut in macOS Chrome changed recently, or you might be thinking of the shortcut in Firefox. In which case, if Opt+Cmd+L is the shortcut in Safari on macOS, it could be allowable to set that as an alternate for "consistency" between browsers (like the Shift+Cmd+N remap for Firefox to open Private windows with the same shortcut as other macOS web browsers). But the correct shortcut specifically for Chrome on macOS (Shift+Cmd+J) should also be set independently. 

Thanks for submitting these, I'll have to add some of them to my own Kinto-based project (Toshy).

PR #866, opened 2024-06-05
'--break-system-packages' not Working in Older Versions of Ubuntu

vshallc:
ref: https://stackoverflow.com/questions/78357791/how-to-apply-break-system-packages-conditionally-only-when-the-system-pip-py


ashirazi2:
Can confirm this fixes install for me on Ubuntu 22.04.5 LTS

PR #769, opened 2022-11-10
New text input field ClassNN for control panels

RedBearAK:
Older Windows control panel windows are also instances of explorer.exe. This adds a new known text input field "focused control" ClassNN to the criteria that controls when to avoid blocking or remapping Enter/Delete/Backspace keys. 

Update to Finder Mods section of `kinto.ahk`.

PR #768, opened 2022-11-10
[Win11] Tab nav with Cmd+Shift+Braces in Explorer

RedBearAK:
Update 22H2 for Windows 11 has brought back tabs in Windows Explorer (the file manager). The tabs use Ctrl+Shift+Tab/Ctrl+Tab for tab navigation.

 This PR enables tab nav with Cmd+Shift+Braces (Ctrl+Shift+Braces) as an update to the Finder Mods section of `kinto.ahk`.

PR #867, opened 2024-06-11
[Win] New Fixes for Finder Mods (2024-06)

RedBearAK:
@rbreaves 

Finder Mods for Windows Explorer had a number of issues that needed to be fixed, primarily coming from how the comparison was being made between the `ClassNN` attribute and a list of different possible strings to match on, for certain keys and shortcuts. 

All issues with Enter, Backspace, Cmd+Backspace (Delete) and the Delete key seem to have been resolved in my testing on Windows 11. Renaming with Enter works on the desktop, and Cmd+Backspace is properly blocked to prevent deleting files on the desktop. Tab navigation shortcuts still working.

### DIFF
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -259,7 +259,7 @@ GroupAdd, intellij, ahk_exe idea64.exe
                 Return
             }
         ; Send Enter to open things while using Cmd+Down in a Control Panel window
-        If ContainsAny(fc, "DirectUIHWND3") {
+        If ContainsAny(fc, "DirectUIHWND3", "SysListView321") {
             Send {Enter}
             Return
         }
--- a/windows/kinto.ahk
+++ b/windows/kinto.ahk
@@ -251,11 +251,18 @@ GroupAdd, intellij, ahk_exe idea64.exe
 
     ^Down::                     ; Cmd-Down: Navigate into the selected directory (has some odd behavior without the conditional)
     For window in ComObjCreate("Shell.Application").Windows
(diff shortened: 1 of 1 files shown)


RedBearAK:
@rbreaves 

And of course after jumping into a different account on the same Windows 11 machine, I ran into a few more issues, and inconsistent `ClassNN` naming of different sections of the Explorer window. But I put in some additional logic to keep an error dialog from appearing in an old Control Panel window (Fonts), which is still part of Explorer. 

It looks like I can't completely block the accidental usage of the Delete key in the desired places, but at least the fix for accidentally deleting a bunch of files on the desktop with Cmd+Backspace should still be working. 

This PR should fix the same problem mentioned in issue #787, that was also something targeted by PR #795. I think I'll close that older PR, and also PR #769. They are obsoleted by this PR. 

Edit: And also PR #768 is included in this PR, so closed that too.

RedBearAK:
@rbreaves 

Added another minor tweak that tries to get Cmd+Down to work on the desktop, for opening files and folders in place of the Enter key. 

I've noticed a lot of strange inconsistency in both Explorer windows and the desktop, where sometimes things just don't work as intended right away, but the later they do, as if the `ClassNN` is different at first and doesn't match. Or perhaps there is some other cause related to restarting the AutoHotkey script to test the changes. 

But once it starts working, it seems to keep working after that. 

I'm not aware of anything else I can improve upon with this PR at this time, and I'm done updating this Win11 machine, so I probably won't mess with this again.

PR #885, opened 2024-11-12
feat: add cursor to vscode string patterns

stefvw93:
Adds support for [Cursor](https://www.cursor.com/), which is a fork of VS Code with AI assistant integrations.

### DIFF
--- a/linux/kinto.py
+++ b/linux/kinto.py
@@ -41,7 +41,7 @@ terminals = [
 terminals = [term.casefold() for term in terminals]
 termStr = "|".join(str('^'+x+'$') for x in terminals)
 
-mscodes = ["code","vscodium"]
+mscodes = ["code","vscodium","cursor"]
 codeStr = "|".join(str('^'+x+'$') for x in mscodes)
 
 sublimes   = ["Sublime_text","subl"]


jonatas-ext:
Hi. How is this issue going? I’ve also started using Cursor recently, and it’s been frustrating. It would be great if it supported all IDEs based on VS Code. Support for the online versions would also be helpful. Maybe a button or a keyboard shortcut to activate 'VS Code mode' would be enough.

jonatas-ext:
UP!

RedBearAK:
@jonatas-ext 

I think your other comment got lost, it only shows up in my email thread. I'm not the Kinto dev, although I have contributed to Kinto in the past. Now I maintain an alternative to Kinto that I called Toshy: 

https://github.com/RedBearAK/toshy/

Its most obvious advantage over Kinto is that the keymapper fork has a number of fixes and supports various Wayland environments, while Kinto still only works in X11/Xorg sessions. But among other differences the default config file in Toshy also has many updates from Kinto's config, including a more extensive list of VSCode variants, and I can add more quite easily: 

```py
vscodes = [
    "code",
    "code - oss",
    "code-oss",
    "cursor",           # New VSCode variant with A.I. 
    "vscodium",
]
vscodes                         = [x.casefold() for x in vscodes]
vscodeStr                       = toRgxStr(vscodes)
```

It at least supports Cursor already.

You said in your email: _"Support for the online versions would also be helpful."_ Toshy's keymapper fork also has support for matching on the window name in addition to the application class, so it might be possible in Toshy to match on a tab name in a web browser to enable activating the VSCode remaps while technically being in the same browser app window. Toshy's config fixes `Cmd+W` behavior with a selection of different "dialogs" by checking the window name.

Check it out if you're interested. You don't necessarily need to uninstall Kinto, but you will need to at least stop the Kinto/xkeysnail service and disable Kinto's autostart feature in the GUI app or tray icon menu, to prevent Kinto from auto-loading at startup, or there will be a conflict over the `uinput` device that would prevent Toshy from working. Otherwise Kinto and Toshy install and operate very differently. 

If you install Toshy and have a question or issue, submit it on the Toshy repo: 

https://github.com/RedBearAK/toshy/issues/

PR #898, opened 2025-04-18
[Ticket 1] Exclude VS Code from key remapping by Kinto

t-tang-rfc:
- No super/ctrl swap
- No keymap def

Tested on Ubuntu 20.4 LTS desktop env.


t-tang-rfc:
Sorry, this pull request is meant for a fork.
It should not go to the upstream.

